PR #15, opened 2019-07-01
Change travis build to use OpenJDK instead of Oracle

j-lum:
All builds with references to Oracle JDK11 breaks as Oracle appears
to be restricting headless downloads. Travis is also moving to remove
support for Oracle JDKs from install-jdk.sh

No work is required by students as their local builds on Oracle JDK
should behave similarly to OpenJDK.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


pyokagan:
@j-lum Please open a PR on AB-4 instead.

PR #20, opened 2019-07-18
Tutorial: Port code for RemarkCommand

j-lum:
Now that the code base for AB3 has stabilized, let's port the removed tutorial from AB4 to AB3.
This should belong on its own branch. 

This commit does not contain the tutorial adoc. I'll be opening a PR to merge that into master.

### DIFF
--- a/src/main/java/seedu/address/logic/commands/RemarkCommand.java
+++ b/src/main/java/seedu/address/logic/commands/RemarkCommand.java
@@ -1,0 +1,16 @@
+package seedu.address.logic.commands;
+
+import seedu.address.model.Model;
+
+/**
+ * Changes the remark of an existing person in the address book.
+ */
+public class RemarkCommand extends Command {
+
+    public static final String COMMAND_WORD = "remark";
+
+    @Override
+    public CommandResult execute(Model model) {
+        return new CommandResult("Hello from remark");
+    }
+}
--- a/src/main/java/seedu/address/logic/commands/RemarkCommand.java
+++ b/src/main/java/seedu/address/logic/commands/RemarkCommand.java
@@ -6,6 +6,7 @@ import static seedu.address.logic.parser.CliSyntax.PREFIX_REMARK;
 import seedu.address.commons.core.index.Index;
 import seedu.address.logic.commands.exceptions.CommandException;
 import seedu.address.model.Model;
+import seedu.address.model.person.Remark;
 
 /**
  * Changes the remark of an existing person in the address book.
@@ -25,13 +26,13 @@ public class RemarkCommand extends Command {
     public static final String MESSAGE_ARGUMENTS = "Index: %1$d, Remark: %2$s";
 
     private final Index index;
-    private final String remark;
+    private final Remark remark;
 
     /**
      * @param index of the person in the filtered person list to edit the remark
      * @param remark of the person to be updated to
      */
-    public RemarkCommand(Index index, String remark) {
+    public RemarkCommand(Index index, Remark remark) {
         requireAllNonNull(index, remark);
 
         this.index = index;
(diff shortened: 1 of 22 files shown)


damithc:
I wonder if the initial steps can be even smaller. e.g., start with a simpler form of the command `remark` (not `remark 2 r/ good`). This will result in a smaller first commit.

On a related note, the current breakdown still feels like written by someone who knows the model solution, whereas the reader doesn't have that kind of expertise. Hence, the steps given may sound 'magical' to the reader (i.e., there is an empathy gap between the writer and the reader, which makes it harder for the reader to connect with the content). 
Keep in mind that we are not only training students to work with AB-3 code base, but also training them _how to use a strategy of 'tiny increments' when working in an unfamiliar code base_. The latter has more value but requires more effort in writing. For example, we should also explain the thought process in a way that can be generalized (`trace through a part that is similar to our target feature` instead of `trace through the delete command`)

j-lum:
From the code, we start from 
1. recognizing `remark` and ignoring everything else
1. recognizing the whole argument string
1. parsing the string to obtain more useful data

So I don't think the code needs to change, maybe we can emphasize the build up a bit more in the tutorial?

damithc:
> So I don't think the code needs to change, maybe we can emphasize the build up a bit more in the tutorial?

The first commit has more code than necessary. e.g.,
* any mention of prefix
* an elaborate help message
* test case

These are things an experienced AB programmer would include in the first step (because she knows these will be needed eventually) but not a newcomer.

j-lum:
I'm writing a new bridging section between the intro and the first bits to code to help the students case out the problem and break it down into logical components. From there we can re-evaluate how we want to structure the commits since the old tutorial (and its code) was meant to be the model answer.

j-lum:
Added a new intermediate step before `Logic: Teach the app to accept 'remark' but do nothing `.

The rationale was that students with little exposure to the code base isn't going to write standard-compliant code right from the code and we should ease them into the process.

src/main/java/seedu/address/logic/commands/RemarkCommand.java
```diff
@@ -1,5 +1,8 @@
 package seedu.address.logic.commands;
 
+import static seedu.address.logic.parser.CliSyntax.PREFIX_REMARK;
```
damithc, line 4: Not needed for this step

src/main/java/seedu/address/logic/commands/RemarkCommand.java
```diff
@@ -9,8 +12,18 @@
 
     public static final String COMMAND_WORD = "remark";
 
+    public static final String MESSAGE_USAGE = COMMAND_WORD + ": Edits the remark of the person identified "
+            + "by the index number used in the last person listing. "
+            + "Existing remark will be overwritten by the input.\n"
+            + "Parameters: INDEX (must be a positive integer) "
+            + PREFIX_REMARK + "[REMARK]\n"
+            + "Example: " + COMMAND_WORD + " 1 "
+            + PREFIX_REMARK + "Likes to swim.";
```
damithc, line 29: not needed for this step

damithc (changes requested):
2nd step needs to be simplified further as it contains things the reader will not see the need for. Plus, it needs to match the first step: currently, the first step creates a Result object and the second step switches to the Exception route.

damithc:
Is there a 1-1 mapping between the commits here and the sections in the tutorial? It's ideal if there is. In that case, section names and commit subjects should match closely.

damithc:
Let's move to the next iteration

j-lum:
Not all commit titles are strict 1-1 matches but they are close enough with some minor differences in grammar. 

Going to shove this through for tutorial on week 6.

PR #22, opened 2019-07-18
Tutorial: Removing field

j-lum:
In lieu of a tutorial on how to morph which can be hard to write and maintain,
we provide a simpler and shorter tutorial on how to remove a field instead.

Students trying to morph the product can repeat this tutorial to get the model
to a state where they can work with.


docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+:note-caption: :information_source:
+:warning-caption: :warning:
+:source-highlighter: highlightjs
+endif::[]
+
+[quote, Charles Maturin, Motto to "Bertram"]
+'Tis well to be merry and wise,
+'Tis well to be honest and true;
+'Tis well to be off with the old love,
+Before you are on with the new.
```
damithc: Let's limit to SE quotes only.

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+In this tutorial, we'll do exactly just that and remove the `Address` field from `Person`.
+
+== Safely deleting `Address`
+
+Fortunately, the IntelliJ IDEA provides a robust refactoring tool that can identify _most_ usages.
+Let's try to use it as much as we can.
+
+=== Assisted refactoring
+We start by identifying ``Address``'s usages.
```
damithc: Explain: What is the rationale behind this step? i.e., what's the connection between the `Address` field and this class?

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+Select the `Address` class and perfom a `Refactor` > `Safe Delete` through the menu.
```
damithc: _select_ where? Mention `right-click`?

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+.Unsafe delete?
```
damithc: Don't get creative with captions 😁   Just state exactly what the image shows.

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+image::UnsafeDelete.png[width=787px. height=238px]
+{empty} +
+
+Choose to `View Usages` and you should be presented with a list of conflicts.
+
+.Short list of conflicts
```
damithc: _short_?

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+image::SafeDeleteConflicts.png[width=955, height=640px]
+{empty} +
+
+Resolve each conflict by recursively performing ``Safe Delete``s.
```
damithc: These terms may be unclear to the reader: _resolve_, _recursively_

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+=== Assisted refactoring
+We start by identifying ``Address``'s usages.
+Select the `Address` class and perfom a `Refactor` > `Safe Delete` through the menu.
+
+.Unsafe delete?
+image::UnsafeDelete.png[width=787px. height=238px]
+{empty} +
+
+Choose to `View Usages` and you should be presented with a list of conflicts.
```
damithc: Explain the meaning of _conflict_ (in this context)?

docs/tutorials/RemovingFields.adoc
damithc: Say something like the below?
>For each conflict, do the following:
>1. ...

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+. Select `Yes` when prompted to remove getters and setters
+. Select `View Usages` again image:UnsafeDeleteOnField.png[width=1145px, height=583px]
+. Remove the usages of `address` and select `Do refactor` when you are done.
+. Repeat the steps for the remaining usages of `Address`
+
+After you are done, verify that the application still works by compiling and running it again.
+
+=== Manual refactoring
+
+Unfortunately, there are usages of `Address` that IntelliJ IDEA cannot identify. Let's fix them.
```
damithc: How to find such leftovers? Do a text search? Given we are removing a person field, the reader should be looking through the fxml files anyway, right?

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+You may have noticed that there is a `$address` in each `PersonCard`.
```
damithc: Unlikely to notice unless looking for it :-p

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+After removing the `Label`, we can proceed to formally test our code.
+If everything went well, you should have most of your tests pass.
+Fix any remaining errors until the tests all pass.
+
+== Tidying up
+
+At this point, your application is working as intended and all your tests are passing.
+What's left to do is to clean up references to `Address` in test data and documentation.
+
+A simple way to generate valid test data is to delete the storage file `data/addressbook.json` and use your application to add dummy values.
```
damithc: There are small links missing in the explanation which could leave the reader puzzled. For example, at this point the reader might wonder:
* Why there is a need to clean up tests if tests are passing? 
* What _test data_ are we talking about? Why the need to generate a _valid_ test data? How did it become invalid?

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,97 @@
...
+Don't forget to modify a copy of valid data to produce invalid data!
```
damithc: This requires a lot of background knowledge to understand.

docs/tutorials/RemovingFields.adoc
damithc: We can list places where there could be left-over references e.g., fxml files, test data files, documentation files, ...

damithc (changes requested):
The overall picture is there 👍 . Just a bunch of small holes that need to be filled up. Don't assume too much background knowledge on the part of the reader.

docs/tutorials/RemovingFields.adoc
j-lum: Stole a quote from your site.

docs/tutorials/RemovingFields.adoc
j-lum: Added a bit to distinguish between the `address` member in `Person` and the `Address` class.

docs/tutorials/RemovingFields.adoc
j-lum: A bit hard to reword resolve but let's see if the new wording works.

docs/tutorials/RemovingFields.adoc
j-lum: Was supposed to be shortlist instead of short list. Let's just change it to list then.

docs/tutorials/RemovingFields.adoc
j-lum: Changed select to right click and sneaked in a typo fix on `perform`.

docs/tutorials/RemovingFields.adoc
j-lum: Changed it to `Safe Delete Conflicts` as shown in the screenshot and explained the meaning of SDCs.

docs/tutorials/RemovingFields.adoc
j-lum: We actually go through an example later on, but I've added some explanations and moved the `recursive` bit from above down here.

docs/tutorials/RemovingFields.adoc
j-lum: It's in the screenshot, not the fxml file but I've changed the wording a bit.

docs/tutorials/RemovingFields.adoc
j-lum: Clarified this part a bit.

AyushChatto:
Hi I was trying to run through the steps of the tutorial, and these lines:
```
4. Remove the usages of address and select Do refactor when you are done.

5. Repeat the steps for the remaining usages of Address
```

are a bit ambiguous, because the usages are within constructors etc. Perhaps we should mention that they can be replaced with dummy `new Address()` instead? Or is there some other method to remove the setter and getter functions.

The usage I'm referring to is as such:

```
Address updatedAddress = editPersonDescriptor.getAddress().orElse(personToEdit.getAddress());

.
.
.

return new Person(updatedName, updatedPhone, updatedEmail, updatedAddress, updatedTags);
```

damithc:
@j-lum can go ahead with the next iteration

j-lum:
Basically remove means `delete` in most cases and will most likely break code in multiple places. This might be uncomfortable for some students so I've added a small aside to let them know that this is expected behavior and it's up to them to fix up the code.

damithc (approved):

damithc:
@pyokagan have time to do a quick check?

pyokagan:
> @pyokagan have time to do a quick check?

Not really at this moment, but can see if any of my comments in my previous reviews apply here.

docs/tutorials/RemovingFields.adoc
```diff
@@ -0,0 +1,125 @@
...
+:toc: macro
+:toc-title:
+:imagesDir: ../images/remove
+:stylesDir: ../stylesheets
+:xrefstyle: full
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:warning-caption: :warning:
+:source-highlighter: highlightjs
```
pyokagan: e.g. this

j-lum: Yeah I'm going through it again with a fine-comb. Fixing a lot of old problems.

j-lum:
Shoving this through because it's blocking the week 6 tutorial.

PR #23, opened 2019-07-18
Tutorial: Tracing code

j-lum:
Feedback from previous students indicated that they have some problem grokking
the code base. Let's address this by having a simple tutorial where students
learn how to apply the tools at hand to trace the execution flow of code.


j-lum:
Moved images into the images folder: 
https://deploy-preview-23--addressbook-level35.netlify.com/tutorials/tracingcode

docs/tutorials/TracingCode.adoc
```diff
@@ -0,0 +1,270 @@
...
+
+As you know, the first step of debugging is to put in a breakpoint where you want the debugger to pause the execution.
+For example, if you are trying to understand how the App starts up, you would put a breakpoint in the first statement
+of the `main` method. In our case, we would want to begin the tracing at the very point where the App start processing
+user input i.e., somewhere in the UI component, and then trace through how the execution proceeds through the UI component.
+However, that the execution path through a GUI is often somewhat obscure due to various _even-driven mechanisms_
+used by GUI frameworks, which happens to be the case here too. Therefore, let us put the breakpoint where the UI transfers
+control to the Logic component. According to the sequence diagram, the UI component yields control to the Logic component
+through a method named `execute`. Searching through the code base for `execute()` yields a promising candidate in
+`seedu.address.ui.CommandBox.CommandExecutor`.
```
damithc, line 52: >However, ~~that~~ the execution path through

docs/tutorials/TracingCode.adoc
```diff
@@ -0,0 +1,270 @@
...
+[a/ADDRESS] [t/TAG]...` For this tutorial we will be issuing the command `edit 1 n/Alice Yeoh`.
+
+[TIP]
+.Taking notes
+====
+Over the course of the debugging session, you will encounter every major component in the application. Try to jot down
+what happens inside the component and where the execution transfers to another component.
+====
+
+. To start the debugging session, simply `Run` > `Debug `Main`
```
damithc: There is an extra back-tick in the rendered view.

docs/tutorials/TracingCode.adoc
```diff
@@ -0,0 +1,270 @@
...
+     try {
+         //We can deduce that the previous line of code modifies model in some way since it's being stored here.
+         storage.saveAddressBook(model.getAddressBook());
+     } catch (IOException ioe) {
+         throw new CommandException(FILE_OPS_ERROR_MESSAGE + ioe, ioe);
+     }
+
+     return commandResult;
+ }
+----
```
damithc, line 120: Code overflowing to the next line

docs/tutorials/TracingCode.adoc
```diff
@@ -0,0 +1,270 @@
...
+`editPersonDesciptor` through the `Variable` tool window.
++
+image:EditCommand.png[width=666px, height=313px]
+. Let's continue stepping through until we return to `LogicManager#execute()`.
++
+The sequence diagram below shows the details of the execution path through the Logic component.
+Does the execution path you traced in the code so far matches with the diagram?
++
+.Tracing an `edit` command through the Logic component
+image::LogicSequenceDiagram.png[width=1080px, height=476px]
```
damithc: return arrows should be dashed

damithc: Also, `[command: EditCommand]`

damithc: If possible, show the full method call e.g., `parse(cmdString)`. Otherwise it is hard to distinguish between method names, parameters, return values.

damithc (changes requested):
Some nits

docs/tutorials/TracingCode.adoc
j-lum: Removed the extra word.

docs/tutorials/TracingCode.adoc
j-lum: Removed.

docs/tutorials/TracingCode.adoc
j-lum: Anything over 120 overflows when the screen is maximized, fixed all code examples.

docs/tutorials/TracingCode.adoc
j-lum: Return arrows dashed, named instance of EditCommand, changed function calls to use real parameter names.

damithc (approved):

damithc:
I'm OK with the current version.Thanks for the hard work @j-lum 
@pyokagan any inputs? It's OK if you don't have time for a full review. Just let us know if you notice any problems that we missed.

pyokagan:
Review with an eye for maintenance:

* `ComponentDiagram.png` and `ArchitectureSequence.png` are mentioned in the tutorial to be the same ones as the developer guide. However, they are new PlantUML diagrams that are currently *not* used in the Developer guide. If we are switching to PlantUML, there should be a PR before that explicitly switches everything to plantuml beforehand.

* In the same vein, it seems the `docs/images/reading/` directory is set aside for images that are only used for this tutorial (By the way, why "reading" when the tutorial is called `TracingCode.adoc`?). Since `ComponentDiagram.png` and `ArchitectureSequence.png` come from the developer guide, they should be in `docs/images/` or `docs/images/DeveloperGuide/`?

* Please name the PlantUML files so that they have the same filename as their output images, so it is immediately obvious which generates what. e.g. `docs/tutorials/uml/architecture-sequence-diagram.puml` should be `docs/tutorials/uml/ArchitectureSequence.puml`. By the way, if `component-diagram.puml` is supposed to be used in the developer guide, then it shouldn't be in the `docs/<<<<tutorials>>>>/uml` directory?

* The screenshots reference `addressbook-level35`. Given that the repo will be renamed to `addressbook-level3` in the future it would be better to make it `addressbook-level3` now. (We could keep it as `addressbook-level35`, but it would not be worth the complaints about the inconsistency with other ABs in the future)

docs/tutorials/TracingCode.adoc
```diff
@@ -0,0 +1,274 @@
...
+:toc: macro
+:toc-title:
+:imagesDir: ../images/reading
+:stylesDir: ../stylesheets
+:xrefstyle: full
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:warning-caption: :warning:
+:source-highlighter: highlightjs
```
pyokagan: Why must this be set to `highlightjs` on GitHub?

docs/tutorials/TracingCode.adoc
```diff
@@ -0,0 +1,274 @@
...
+{empty} +
+
+Bingo! `MainWindow#executeCommand()` seems to be exactly what we're looking for!
+
+Now let's set the breakpoint. First, double-click the item to reach the corresponding code. Once there, click on the left
+gutter to set a breakpoint, as shown below.
+image:LeftGutter.png[width=1171px, height=154px]
+
+
+[qanda]
```
pyokagan: ?

pyokagan:
> The screenshots reference addressbook-level35. Given that the repo will be renamed to addressbook-level3 in the future it would be better to make it addressbook-level3 now.

I'd also be fine with leaving the screenshots as they are actually. It may not be a problem in practice. Might be a good idea use the correct name when making screenshots for later tutorials though.

j-lum:
>     * `ComponentDiagram.png` and `ArchitectureSequence.png` are mentioned in the tutorial to be the same ones as the developer guide. However, they are new PlantUML diagrams that are currently _not_ used in the Developer guide. If we are switching to PlantUML, there should be a PR before that explicitly switches everything to plantuml beforehand.
> 
>     * In the same vein, it seems the `docs/images/reading/` directory is set aside for images that are only used for this tutorial (By the way, why "reading" when the tutorial is called `TracingCode.adoc`?). Since `ComponentDiagram.png` and `ArchitectureSequence.png` come from the developer guide, they should be in `docs/images/` or `docs/images/DeveloperGuide/`?

Filing a PR to update AB-4 to use PlantUML diagrams. When it gets backported I'll update the tutorial to pull images from the right place. The old tutorial was supposed to be 'code reading' but it got changed during the editing process. Will update the directory name to match.

>     * Please name the PlantUML files so that they have the same filename as their output images, so it is immediately obvious which generates what. e.g. `docs/tutorials/uml/architecture-sequence-diagram.puml` should be `docs/tutorials/uml/ArchitectureSequence.puml`. By the way, if `component-diagram.puml` is supposed to be used in the developer guide, then it shouldn't be in the `docs/<<<<tutorials>>>>/uml` directory?
Will update the naming convention to match. Directory stuff should resolve after the PR from 4 gets backported.

>     * The screenshots reference `addressbook-level35`. Given that the repo will be renamed to `addressbook-level3` in the future it would be better to make it `addressbook-level3` now. (We could keep it as `addressbook-level35`, but it would not be worth the complaints about the inconsistency with other ABs in the future)

When we pull the plug on AB-3 and rename this repo to 3 I'll make the change.

j-lum:
@pyokagan Think we are ready to move forward with the set of tutorials for ab-3 now. Please take a look when you have some time.

docs/tutorials/TracingCode.adoc
```diff
@@ -0,0 +1,269 @@
+= Tutorial - Tracing a Command Execution Path
+:site-section: DeveloperGuide
+:toc: macro
```
pyokagan: I noticed that the toc macro isn't used. Do you intend to have a toc? If no, can remove the toc-related attributes.

docs/tutorials/TracingCode.adoc
```diff
@@ -0,0 +1,269 @@
...
+[source, java]
+----
+public void setFeedbackToUser(String feedbackToUser) {
+    requireNonNull(feedbackToUser);
+    resultDisplay.setText(feedbackToUser);
+}
+----
+. Finally, we step through until we reach the end of `MainWindow#executeCommand()`.
+
+[discrete]
```
pyokagan: I suspect this was left around during the toc days. Can be removed if so.

docs/tutorials/TracingCode.adoc
```diff
@@ -0,0 +1,269 @@
...
+
+In this tutorial, we traced a valid edit command from raw user input to the result being displayed to the user. From
+this tutorial, you learned more about the inner workings of AddressBook and how the various components mesh together
+to form one cohesive product.
+
+
+{empty} +
+[lead]
+.What's next?
+====
```
pyokagan: This renders as an "example": https://deploy-preview-23--addressbook-level35.netlify.com/tutorials/tracingcode

docs/tutorials/TracingCode.adoc
```diff
@@ -0,0 +1,269 @@
...
+[discrete]
+== Conclusion
+
+In this tutorial, we traced a valid edit command from raw user input to the result being displayed to the user. From
+this tutorial, you learned more about the inner workings of AddressBook and how the various components mesh together
+to form one cohesive product.
+
+
+{empty} +
+[lead]
```
pyokagan: Does the `[lead]` do anything? I thought it only applied to paragraphs.

docs/tutorials/uml/ComponentDiagram.puml
```diff
@@ -0,0 +1,40 @@
...
+Main -> UI
+
+Logs -up- Commons
+Hidden .down.> Commons
+Hidden .down.> Commons
+Hidden .down.> Commons
+
+Storage .right.>File
+
+hide circle
```
pyokagan: This would not be here if we applied DRY and used the ArchitectureDiagram.puml that's in master ;-)

Anyway I think this file is out of date.

pyokagan (changes requested):
> ```
> Tutorial: Tracing code
> ```

Commit subjects should be imperative, so there needs to be a verb inside
there like "Add tutorial on tracing code" or something.

> ```
> Feedback from previous students indicated that they have some problem grokking
> the code base. Let's address this by having a simple tutorial where students
> learn how to apply the tools at hand to trace the execution flow of code.
> ```

Lines exceed 72 chars. Let's fix this problem once and for all.

In the spirit of DRY, maybe we should not have extra copies of
ComponentDiagram.puml (ArchitectureDiagram.puml) and
LogicSequenceDiagram.puml (ArchitectureSequenceDiagram.puml)? (And
likewise, ComponentDiagram.png and LogicSequenceDiagram.png?)

I'm not a great fan of using `{empty}+` for adding vertical spacing
since it obscures the content of the document for a reader reading the
adoc file. (Sort of like how using \<\p> tags in a HTML file all over the
place makes it hard to read/write documents, which was why markup
formats like Markdown were invented).
Seems like something that CSS should be solving instead. Just an
opinion, not a reject-worthy deal.

pyokagan:
Hrm, I think we need to regenerate the plantuml diagrams since we are using Arial now or something as well.

j-lum:
I'll finish up the Duke stuff and get this out ASAP.

JiaHaoLim:
After Figure 1: 
image
Should be "components" ?

Setting a break point:
image
Should be "event-driven".

Step 20 of Tracing the execution path:
image
Perhaps this step could be clearer by bringing them through individual stepping ins, instead of simply "repeatedly stepping".

Conclusion:
This is a good guide(most steps are clear and easy to follow) for tracing code and debugging, but I'm not sure if it's enough to help them understand the code base. Throughout the tutorial, it shows them the code "hierarchy" but not why. Perhaps to facilitate learning, additional explanations can be given in the same fashion as Step 22.

junrong09:
The `Find Usages` command is extremely useful for developers to dive into a fresh set of code. Great to see that in the guide. 

It might also be helpful to focus on static code analysis by introducing a secondary option to trace through the call stacks without running the debug mode. IntelliJ's `Declaration` and `Find Usages` commands via the `Ctrl+B`/`Ctrl-right click` should provide a swift alternative for developers to get started with the code at their component of interest (e.g. Logic.java).

j-lum:
@junrong09 We do cover `Find Usages` in the tutorial and `Declaration` should be part of the main coursework in the textbook (iirc from teaching last sem).

@JiaHaoLim 
About the last point, there was actually a huge refactor of the documentation (tutorials, how-tos, discussions and reference) but due to time constraints we can only finish tutorials. I have some discussion pieces (where the rationale and stuff should go) sitting on another repo pending approval from Prof before I release them to students. 

Also, it's a bit hard to dump information about the design of AB because the principles are not discussed until much later.

docs/tutorials/TracingCode.adoc
pyokagan: > I noticed that the toc macro isn't used. Do you intend to have a toc? If no, can remove the toc-related attributes.

pyokagan (changes requested):
OK, just one thing which I think you missed out.

> ```
> Add tutorial on tracing code
>
> Feedback from previous students indicated that
> they have some problem grokking the code base.
> Let's address this by having a simple tutorial
> where students learn how to apply the tools at
> hand to trace the execution flow of code.
> ```

By the way, the text wrapping seems pretty narrow ;-)

j-lum:
Whoops fixed that.

pyokagan (approved):

damithc:
Guys, shall we merge this one?

j-lum:
Do we want to keep all the tutorials on a separate branch or just merge it into master?

damithc:
> Do we want to keep all the tutorials on a separate branch or just merge it into master?

Tutorials can be on `master`, except the `remark` code that we should not merge.

j-lum:
Thanks for all the help reviewing guys! 

A few more tutorials to go...

PR #24, opened 2019-07-18
Tutorial: Implementing the Remark Command

j-lum:
Since we removed the tutorial for the RemarkCommand from AB-4,
let's add a new tutorial for it at the AB-3.5 level.


damithc:
Was there any significant changes from the AB4 version?
I've set up Netlify. But it will trigger only at the next update to the PR. Do a dummy update?

j-lum:
This tutorial is written from scratch so we will have to send it through the review process. Let me push something real quick.

damithc:
hmm... looks like Netlify only captures new PRs created after enabling Netlify :-p

j-lum:
https://deploy-preview-24--addressbook-level35.netlify.com/tutorials/remark
It worked!

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,360 @@
+= AddressBook Level 3 - Tutorial
+:site-section: Tutorial
```
pyokagan: `site-section` is used to determine which entry to highlight in the site navigation bar:

https://github.com/se-edu/addressbook-level35/blob/9aaee4dbfdb9f2ed8a421e556c652abe2afd3fca/docs/templates/_header.html.slim#L30-L41

So there's not much point in setting it to "Tutorial" if there is no corresponding entry there. You can: (1) Not set it at all, which is better than using an unrecognized value, (2) Add an entry on it to the site navigation bar (make sure to update the `site-section` docs!) or (3) Use an existing entry in the navigation bar

j-lum: I actually wrote some ruby to dump them into a `Tutorial` section. Think that got shafted when it broke something. Let me just remove it for now until I can remember what broke. iirc is_root? or function like that got removed in a later version of asciidoc...

j-lum: There was a period where I was struggling to get us up to asciidoctor 2.2 for asciidoctor-diagram to work properly but it ended up being a complete quagmire. I'll get around to it eventually(tm).

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,360 @@
...
+`remark 2 r/Likes baseball`.
+
+We'll assume that you have already set up the development environment as outlined in the Developer's Guide.
+
+=== Create a new `Command`
+
+Looking in the `logic.command` package, you will notice that each existing command have their own class. All the
+commands implement `Command` which means that they must override `execute()`. By convention, `Command` returns an
+instance of `CommandResult` upon success and throws a `CommandException` otherwise. This exception is handled
+gracefully and the error message provided to `CommandException` is printed the application.
```
damithc: The last sentence doesn't read properly.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,360 @@
...
+
+.UML diagram describing the relationship between RemarkCommand and Command
+image::CommandInterface.png[width=385px, height=385px]
```
damithc: Remove spots `I` and `C`

damithc: Shouldn't `Command` be abstract?

damithc: Also, use standard notations to indicate visibility `+`, `-`, ...

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,360 @@
...
+            + PREFIX_REMARK + "Likes to swim.";
+
+    public static final String MESSAGE_NOT_IMPLEMENTED_YET = "Remark command not implemented yet";
+
+    @Override
+    public CommandResult execute(Model model) throws CommandException {
+        throw new CommandException(MESSAGE_NOT_IMPLEMENTED_YET);
+    }
+}
+----
```
damithc: Might as well mention that the reader should copy-paste-modify the code from another command. it's unlikely anybody would type all this from scratch.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,360 @@
...
+            + "Parameters: INDEX (must be a positive integer) "
+            + PREFIX_REMARK + "[REMARK]\n"
+            + "Example: " + COMMAND_WORD + " 1 "
+            + PREFIX_REMARK + "Likes to swim.";
+
+    public static final String MESSAGE_NOT_IMPLEMENTED_YET = "Remark command not implemented yet";
+
+    @Override
+    public CommandResult execute(Model model) throws CommandException {
+        throw new CommandException(MESSAGE_NOT_IMPLEMENTED_YET);
```
damithc: Inline the string. No benefit from defining as a constant.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,360 @@
...
+
+    public static final String COMMAND_WORD = "remark";
+
+    public static final String MESSAGE_USAGE = COMMAND_WORD + ": Edits the remark of the person identified "
+            + "by the index number used in the last person listing. "
+            + "Existing remark will be overwritten by the input.\n"
+            + "Parameters: INDEX (must be a positive integer) "
+            + PREFIX_REMARK + "[REMARK]\n"
+            + "Example: " + COMMAND_WORD + " 1 "
+            + PREFIX_REMARK + "Likes to swim.";
```
damithc: `PREFIX_REMARK`: where is this coming from? It shouldn't exist at this point.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,360 @@
...
+----
+/**
+ * Tokenizes an arguments string and returns an {@code ArgumentMultimap} object that maps prefixes to their
+ * respective argument values. Only the given prefixes will be recognized in the arguments string.
+ *
+ * @param argsString Arguments string of the form: {@code preamble <prefix>value <prefix>value ...}
+ * @param prefixes   Prefixes to tokenize the arguments string with
+ * @return           ArgumentMultimap object that maps prefixes to their arguments
+ */
+----
```
damithc: Some of the code examples overflow to the next line. Check other places too.
image

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,360 @@
...
+
+If you are stuck, check out the sample
+link:++https://github.com/j-lum/addressbook-level35/commit/c0f8419be4fb59bffafcd73026769459e5cf5c5e#diff
+-fc19ecee89c3732a62fbc8c840250508++[here].
+
+== Adding to the model
+
+Now that we have all the information that we need, let's lay the groundwork for some _persistent_ changes. We acheive
+ that by working with the `Person` model. Each field in a Person is encapsulated. This protects them from illegal
+ access and allows us to tightly couple logic specific to the field.
```
damithc: >Each field in a Person is encapsulated. This protects them from illegal  access and allows us to tightly couple logic specific to the field.

-> Something like this?

>Each field in a Person is implemented as a separate class (e.g., a `Name` object represents the person's name). That means we should add a `Remark` class so that we can use a `Remark` object to represent a remark given to a person.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,360 @@
...
+each person.
+
+Simply add
+```
+@FXML
+private Label remark;
+```
+to
+link:++https://github.com/j-lum/addressbook-level35/commit/06f9897d7088d22689de478c94612f8bcbff2964#diff
+-0c6b6abcfac8c205e075294f25e851fe++[`PersonCard`]. `@FXML` is an annotation that marks a private or protected field and
```
damithc: Try not to break a sentence over a code block.

>Simply add the following to the `foo.fxml`:
```
...
```

damithc (changes requested):
Added some comments. Can proceed to the next iteration.

docs/tutorials/Remark.adoc
j-lum: Grammar mistakes and a run on sentence. A little column A, little column B. Fixed it.

docs/tutorials/Remark.adoc
j-lum: Abstract classes have their names in italics. 
Change icons to characters.

docs/tutorials/Remark.adoc
j-lum: Hmm.. Should review of the code go to the other PR?

docs/tutorials/Remark.adoc
j-lum: Preemptively wrapped code at 80 columns

docs/tutorials/Remark.adoc
j-lum: Added it in

docs/tutorials/Remark.adoc
j-lum: Added a note to remind them to do that.

docs/tutorials/Remark.adoc
j-lum: Joined them.

docs/tutorials/Remark.adoc
j-lum: Added a line telling them that they can copy and paste from the repo.

docs/tutorials/Remark.adoc
damithc: Use the 2nd option. We cannot expect students to write italics in the exam; hence the 2nd option is preferred.
image

docs/tutorials/Remark.adoc
damithc: Can fix in the other PR first, but eventually need to fix here too.

docs/tutorials/Remark.adoc
j-lum: There seems to be no nice way to get this in PlantUML. 
Closest we can get is <<{abstract}>. Does that work?

damithc:
At the start of the tutorial, we can advise the reader to follow the 'trace code' tutorial first to find how a command is executed.

j-lum:
My plan is to wait for all the tutorials to enter master and then have a separate PR a landing page for the tutorials along with some changes to the navbar. Can we wait until then or do we want to make it happen this PR?

damithc:
I expect you've realized the following already, but for the record: 
* In se-edu commit organization, we implement the feature first and then reverse engineer it into a series of commits, each a logical step towards the ideal solution. 
* In contrast, this PR should pretend to not know the ideal solution, and show the steps we would take in search of a solution if we can only see just a bit ahead of where we are standing (because that's what a novice would do). 

i.e., the commit organization in these tutorial PRs will be slightly different from other normal se-edu PRs.

damithc:
> I expect you've realized the following already, but for the record:
> 
> * In se-edu commit organization, we implement the feature first and then reverse engineer it into a series of commits, each a logical step towards the ideal solution.
> * In contrast, this PR should pretend to not know the ideal solution, and show the steps we would take in search of a solution if we can only see just a bit ahead of where we are standing (because that's what a novice would do).
> 
> i.e., the commit organization in these tutorial PRs will be slightly different from other normal se-edu PRs.

Sorry, this comment is meant for the code PR.

damithc:
> My plan is to wait for all the tutorials to enter master and then have a separate PR a landing page for the tutorials along with some changes to the navbar. Can we wait until then or do we want to make it happen this PR?

Sure, can wait. No harm mentioning in this page too. People can land in this page without going through the landing page first.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,363 @@
...
+edit it if required. The command should have the format of `remark INDEX r/REMARK`. An example of the command is
+`remark 2 r/Likes baseball`.
+
+We'll assume that you have already set up the development environment as outlined in the Developer's Guide.
+
+=== Create a new `Command`
+
+Looking in the `logic.command` package, you will notice that each existing command have their own class.
+All the commands implement `Command` which means that they must override `execute()`.
+By convention, `Command` returns an instance of `CommandResult` upon success and throws a `CommandException` otherwise.
```
damithc: I don't think it is right to say `By convention` here. It's a hard requirement rather than a convention.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,363 @@
...
+This exception is handled gracefully -- the error message provided to `CommandException` is shown in the `ResultDisplay`.
+
+.UML diagram describing the relationship between RemarkCommand and Command
+image::CommandInterface.png[width=385px, height=385px]
+{empty} +
```
damithc: Notation doesn't match what we cover in the module
image

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,363 @@
...
+
+=== Parse user input
+
+Now let's move on to writing a parser that will extract the index and remark from the input provided by the user.
+
+Create a `RemarkCommandParser` class in the `seedu.address.logic.parser` package. The class must extend the `Parser`
+interface.
+
+.UML diagram describing the relationship between Parser and RemarkCommandParser
+image::ParserInterface.png[width=386px, height=350px]
```
damithc: image

Notation issues

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,363 @@
...
+:imagesDir: ../images/remark
+:stylesDir: ../stylesheets
+:xrefstyle: full
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:warning-caption: :warning:
+:source-highlighter: highlightjs
+endif::[]
+
```
damithc: Add a table of contents?

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,363 @@
+= AddressBook Level 3 - Tutorial
```
damithc: `Tutorial - Adding a new Command`?

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,363 @@
...
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+:warning-caption: :warning:
+:source-highlighter: highlightjs
+endif::[]
+
+
+== Adding your first command
+
```
damithc: `Adding a skeletal version of the new command`?

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,363 @@
...
+recognize the `remark` keyword. Add the new command to the switch-case and return a new instance of `RemarkCommand`.
+
+You can refer to the changes in this
+link:++https://github.com/j-lum/addressbook-level35/commit/72256eacea79dfcf077f5c7cb6f89819c02f090e#diff-5338391f3f6fbb4022c44add6590b74f++[diff].
+
+=== Run the application
+
+Run `Main#main` and try out your new `RemarkCommand`. If everything went well, you should see something like this:
+
+.Successful failure
```
damithc: Use straight-forward self-explanatory captions

damithc (changes requested):
Some cosmetic changes requested. After fixing them, we can ask reviews from a few tutors.

docs/tutorials/Remark.adoc
j-lum: Changed it to just say "Each `Command`".

docs/tutorials/Remark.adoc
j-lum: Do we have a recommended notation for exceptions and generics?

docs/tutorials/Remark.adoc
j-lum: Added one.

docs/tutorials/Remark.adoc
j-lum: Changed the title to match the tracing code tutorial.

docs/tutorials/Remark.adoc
j-lum: Actually changed that to "Introduction" to accurately describe the section.

docs/tutorials/Remark.adoc
damithc: We don't cover either. Perhaps omit them from the diagram as well?

j-lum:
The UML diagrams might look a little weird (as least compared to the DG) but it's strictly because of PlantUML's limitations as I'm unable to change the font color of the members. Also there's one dirty hack to get the class box to resize properly.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,347 @@
...
+We'll assume that you have already set up the development environment as outlined in the Developer's Guide.
+
+=== Create a new `Command`
+
+Looking in the `logic.command` package, you will notice that each existing command have their own class.
+All the commands implement `Command` which means that they must override `execute()`.
+Each `Command` returns an instance of `CommandResult` upon success and throws a `CommandException` otherwise.
+This exception is handled gracefully -- the error message provided to `CommandException` is shown in the `ResultDisplay`.
+
+.UML diagram describing the relationship between RemarkCommand and Command
```
damithc: UML diagram describing the relationship between RemarkCommand and Command -->
~~UML diagram describing~~ The relationship between `RemarkCommand` and `Command`

There are similar cases in other places.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,347 @@
...
+This command allows users of the AddressBook application to add optional remarks to people in their address book and edit it if required.
+The command should have the format of `remark INDEX r/REMARK`.
+An example of the command is `remark 2 r/Likes baseball`.
+
+We'll assume that you have already set up the development environment as outlined in the Developer's Guide.
+
+=== Create a new `Command`
+
+Looking in the `logic.command` package, you will notice that each existing command have their own class.
+All the commands implement `Command` which means that they must override `execute()`.
```
damithc: implement `Command` --> inherit from the abstract class `Command`

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,347 @@
...
+
+If you are stuck, check out the sample link:https://github.com/j-lum/addressbook-level35/commit/c0f8419be4fb59bffafcd73026769459e5cf5c5e#diff-fc19ecee89c3732a62fbc8c840250508[here].
+
+== Adding to the model
+
+Now that we have all the information that we need, let's lay the groundwork for some _persistent_ changes.
+We acheive that by working with the `Person` model.
+Each field in a Person is implemented as a separate class (e.g. a Name object represents the person's name).
+That means we should add a Remark class so that we can use a Remark object to represent a remark given to a person.
+This protects them from illegal access and allows us to tightly couple logic specific to the field.
```
damithc: ~~This protects them from illegal access and allows us to tightly couple logic specific to the field.~~ 

Unnecessary statement with a confusing claim: tight-coupling is usually used in a negative sense.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,347 @@
...
+Since `PersonCard` displays data from a `Person`, we need to update `Person` to get our `Remark` displayed!
+
+=== Modify `Person`
+
+We change the constructor of `Person` to take a `Remark`.
+We will also need to define new fields accordingly to store our new addition.
+
+=== Update other usages of `Person`
+
+Unfortunately, a change to `Person` will cause other commands to break, let's fix these commands to use the updated `Person`!
```
damithc: Make this statement clearer: are you going to explain how to do that in the sections below or the reader is supposed to do that by themselves.

damithc (changes requested):
some cosmetic changes. In the meantime, I'll ask some tutors to dog-food.

j-lum:
Implemented the changes.

kthSim:
Near the end of the "Create a new Command" section
image
- Could be changed to make it more explicit that they need to make a new Prefix with the value of "r/", in case they overlooked the proper command format. 
- Perhaps include the location of CliSyntax.java and a relevant link to the changes in CliSyntax.java as well? 

For the "Hook RemarkCommand into the application" section:
- Perhaps an additional note that the location that the RemarkCommand case is added to the switch-case in AddressBookParser doesn't matter for the benefit of those less familiar with Java programming?

For the "Make the command accept parameters" section:
- IntelliJ might not automatically import the relevant Index package into RemarkCommand.java. If I could make a suggestion, the tutorial could include a small note instructing students to check that "import seedu.address.commons.core.index.Index;" has been added to their code.
-- Same thing for "import static seedu.address.commons.util.CollectionUtil.requireAllNonNull;"

For the "Adding placeholder elements to the UI" section:
- Perhaps the tutorial could state the location of "PersonCard"and "PersonListCard.fxml" to reduce confusion amongst students.

For the "Modify Person" section:
- Might want to remind students to add a relevant getter function (i.e. getRemark) 
- A reminder to update existing functions with the new Remark field (i.e isSamePerson, toString, hashCode, equals, etc.) might be useful for students as well.

For the "Update other usages of Person" section:
- As a hint for students that may be confused, perhaps we could include a link to the part of the dev guide that showcases how AB3's components work together?
- A reminder to change SampleDataUtil.java might be useful for students as well.

For the "Updating Storage" Section:
- The commit doesn't highlight some necessary changes to PersonBuilder.java (adding a new `withRemark` function, updating the relevant existing functions with Remark)

For the "Automagically generating tests" section:
- Automagically should(?) be Automatically. 
- Commit doesn't show the needed addition of VALID_REMARK_AMY/BOB to CommandTestUtil.java

emer7:
For **Adding to the model** `->` **Make use of `Remark`**:

- Perhaps can also include the link from previous section **Add a new `Remark` class**, specifically this sentence: *"A quick copy-paste and search-replace later, you should have something like [this](https://github.com/j-lum/addressbook-level35/commit/c8e6048f0d81f2b4b7797e838226da823892129b#diff-af2f075d24dfcd333876f0fbce321f25)"*.
This is in case the student is lost and also that commit happen to contains the required change for **Make use of `Remark`** section.

For **Adding placeholder elements to the UI** section:

- I agree with Keith above, it can be quite disorientating to find `PersonCard` and `PersonListCard.fxml`, especially with `PersonListCard.fxml` is not in `java` folder but in `resources`.
- I think `PersonCard.java` will be better instead of `PersonCard (no extension suffix)`, because there is FXML mentioned, student can be confused is `PersonCard` a reqular `.java` file or not.
- Since this section I think can be the most disorientating for student as they are handling not only `java` file but also `fxml` file, perhaps can a note to remind students to not panic and to learn by observing how the code is written and structured.

For **Updating the model** section:

- Remind student to update relevant methods inside `Person.java`. Also might want to explain to them `isSamePerson` method. I don't think having the identical remark with another person makes both of that person the same person and hence need not to be updated.
- This section lacks the commit reference. Perhaps can add it in so this pattern is repeated for each section (increase predictability and assurance for student).

For **Finalizing the UI** section:
- I thought the first person will have "$remark is gone" as a remark, but turns out no. Perhaps somebody would also think the same.
image

For **Putting everything together** `->` **Change `Remark`** section:
- The image given might want to include the relevant command that produces that result in the command bar.

For **Writing tests** `->` **Automagically generating tests** section:
- The image given might indicate to go to test for `RemarkCommand` "constructor" method instead of `RemarkCommand#execute` method.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,396 @@
...
+// ...
+private final Index index;
+private final String remark;
+
+/**
+ * @param index of the person in the filtered person list to edit the remark
+ * @param remark of the person to be updated to
+ */
+public RemarkCommand(Index index, String remark) {
+    requireAllNonNull(index, remark);
```
nbriannl: `requireNonNull`? Is this a typo.

nbriannl: Is not consistent with line 113.

I used `requireNonNull` until I realised later on at the last sections of the exercise to import requireAllNonNull from `seedu.address.commons.util.CollectionUtil.requireAllNonNull;`

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,396 @@
...
+
+=== Make the command accept parameters
+
+We start by modifying the constructor of `RemarkCommand` to accept an `Index` and a `String`.
+While we are at it, let's change the error message to echo the values.
+While this is not a replacement for tests, it is an obvious way to tell if our code is functioning as intended.
+
+[source, java]
+----
+import static java.util.Objects.requireNonNull;
```
nbriannl: Is not consistent with line 125 `requireAllNonNull`

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,396 @@
...
+We achieve that by working with the `Person` model.
+Each field in a Person is implemented as a separate class (e.g. a `Name` object represents the person's name).
+That means we should add a `Remark` class so that we can use a `Remark` object to represent a remark given to a person.
+
+=== Add a new `Remark` class
+
+Create a new `Remark` in `seedu.address.model.person`. Since a `Remark` is a field that is similar to `Address`, we can reuse a significant bit of code.
+
+A quick copy-paste and search-replace later, you should have something like link:https://github.com/j-lum/addressbook-level35/commit/c8e6048f0d81f2b4b7797e838226da823892129b#diff-af2f075d24dfcd333876f0fbce321f25[this].
+
```
nbriannl: TL;DR Overall, this part is fine.

This is a relatively hard part for lazy students who are simply zombie-ing and copy pasting to go through the tutorial. To make a remark class in the model requires someone to recognize that there are no constraints to remarks, hence no need to `checkArguments()` and check an Remark equivalent for `isValidAddress`

I would argue it's some copy and paste, with some required thinking, not necessarily a 'quick' one

Anyways, someone's who's lazy can just check the Github repo and copy and whole Remark.java wholesale, so it's fine overall

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,396 @@
...
+<Label fx:id="remark" styleClass="cell_small_label" text="\$remark" />
+```
+into link:https://github.com/j-lum/addressbook-level35/commit/06f9897d7088d22689de478c94612f8bcbff2964#diff-12580431f55d7880578aa4c16f249e71[`main/resources/view/PersonListCard.fxml`].
+
+That's it! Fire up the application again and you should see something like this:
+
+.$remark shows up in each entry
+image::$Remark.png[]
+
+== Modify `Person` to support a `Remark` field
```
nbriannl: Doing this first thing in the morning after eating breakfast, my **personal** opinion is that **Modify Person to support a Remark field** gotta be the hardest part of the tutorial.

You've been giving commit links for people to reference to, to not provide one all of a sudden is quite jarring. I think the relevant commit is this (please double-check)
https://github.com/j-lum/addressbook-level3/commit/b03b788d5a462403ea7f4dce9a832b62f729f7c4

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,396 @@
...
+=== Modify `Person`
+
+We change the constructor of `Person` to take a `Remark`.
+We will also need to define new fields and accessors accordingly to store our new addition.
+
+=== Update other usages of `Person`
+
+Unfortunately, a change to `Person` will cause other commands to break, you will have to modify these commands to use the updated `Person`!
+
+TIP: Use the `Find Usages` feature in IntelliJ IDEA on the `Person` class to find these commands.
```
nbriannl: Tip only makes me overwhelmed seeing all these

image

What helped me was to run the app and see what broke during compilation 😆

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,396 @@
...
+
+== Updating Storage
+
+AddressBook stores data by serializing `JsonAdaptedPerson` into `json` with the help of an external library -- Jackson.
+Let's update `JsonAdaptedPerson` to work with our new `Person`!
+
+While the changes to code may be minimal, the test data will have to be updated as well.
+
+Check out link:https://github.com/j-lum/addressbook-level35/commit/78b2a7438a9f9d828b7f09e190ff6130790966c0[this commit] to see what the changes entail.
```
nbriannl: `PersonBuilder` changes are split across https://github.com/j-lum/addressbook-level3/commit/b03b788d5a462403ea7f4dce9a832b62f729f7c4 and https://github.com/j-lum/addressbook-level3/commit/78b2a7438a9f9d828b7f09e190ff6130790966c0
The commit link under Updating Storage https://github.com/j-lum/addressbook-level3/commit/78b2a7438a9f9d828b7f09e190ff6130790966c0 includes changes for testing

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,396 @@
...
+----
+public PersonCard(Person person, int displayedIndex) {
+    //...
+    remark.setText(person.getRemark().value);
+}
+----
+
+.The remark label is bound properly!
+image::RemarkBound.png[]
+
```
nbriannl: No mention to add remark field to `/data/addressbook.json`
If you don't do that you can't init the data for AB3 after making changes.
I'm not sure how much we are to spoonfeed them in this tutorial, but just a headsup.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,396 @@
...
+== Putting everything together
+
+After the previous step, we notice a peculiar regression -- we went from displaying something to nothing at all.
+However, this is expected behavior as we have never changed `RemarkCommand` at all!
+
+=== Update `RemarkCommand`
```
nbriannl: For **Update RemarkCommand**
I had to change `RemarkCommandParser` as well.

docs/tutorials/Remark.adoc
```diff
@@ -0,0 +1,396 @@
...
+
+    /**
+     * Generates a command execution success message based on whether the remark is added to or removed from
+     * {@code personToEdit}.
+     */
+    private String generateSuccessMessage(Person personToEdit) {
+        String message = !remark.value.isEmpty() ? MESSAGE_ADD_REMARK_SUCCESS : MESSAGE_DELETE_REMARK_SUCCESS;
+        return String.format(message, personToEdit);
+    }
+----
```
nbriannl: Code block also does not show to implement `equals`
See: https://github.com/j-lum/addressbook-level3/blob/1492fb0d4e8a075f2481028377e0d113e7c01b76/src/main/java/seedu/address/logic/commands/RemarkCommand.java

nbriannl (commented):
I ran through the tutorial. Here are my observations while trying it out.

kylerwsm:
I apologise for the late comments. Since the tutorial is aimed such that it should be easy to follow and gain general understanding of the addressbook, I have some comments regarding how the tutorial may be amelioriated such that students may be able to understand it better.

### Suggestion 1: Addition of Background Context
Screenshot 2019-09-08 at 2 05 58 PM
I observe that there are multiple instances that calls for students to follow a certain standard in implementing a certain algorithm. While the better students may understand why there is a need to abide to such standards, I am skeptical whether the lacking ones will be able to do so. In this case of the `Parser`, for example, there is no explanation given after indicating that it is a *must* to to inherit from it. In my opinion, an instruction without explanation does not really help at all.

### Suggestion 2: Standardisation of Commonly Used Diagrams
Screenshot 2019-09-08 at 2 19 31 PM
I have also noticed disparity amongst diagrams in this tutorial and those taught over tutorial classes. While the diagrams are all class diagrams, they are drawn in vastly different ways that may invoke confusion with students. In worse cases, it might not even come across as the same diagrams to students. I think the diagram can be labelled and explained, which may alleviate these possible confusion, while allowing students to grasp and internalise the key principles when drawing such diagrams.

damithc:
Rename the file to match other tutorials? e.g., `AddingCommands.adoc`?

docs/tutorials/Remark.adoc
j-lum: I have a very nasty feeling some students are going to clone the last step of the tutorial and call it a day. I was thinking of inserting some gotchas to distinguish between students who actually attempted  the tutorial vs just cloning but I decided against it since they are responsible for their own learning I guess. 

I'll pull the 'quick' out and add a tip to make them consider what they need to copy and paste in.

docs/tutorials/Remark.adoc
j-lum: Technically they should have finished the removing address tutorial before this and have some competency navigating this. I have my fingers crossed.

docs/tutorials/Remark.adoc
j-lum: Yeah some parts of the tutorial are really short in text and intense in execution because it's pretty much impossible to guide them through with code snippets. Linked the commit to save them some pain.

docs/tutorials/Remark.adoc
j-lum: Yeah this one is something that we MUST mention. I rather add the line than to respond to 3am emails.

j-lum: Added a line to ask them to delete the file and have a new compliant one generated. Let's not risk them editing JSON by hand.

docs/tutorials/Remark.adoc
j-lum: Added the implementation in a previous step.

j-lum:
> ### Suggestion 1: Addition of Background Context
This should go under a special type of documentation called "discussions". As much as possible we try to avoid adding text that does not explicitly push them towards the completion of the tutorial. We might want to cover this during the actual in-person tutorial?

> ### Suggestion 2: Standardisation of Commonly Used Diagrams
This is a bit icky as this is one of the few UML diagrams that has details required for implementation. How would you suggest changing it? Hopefully from AB-3 onwards they will be dealing with exclusively with PlantUML diagrams but it's a chore to convert all of them and make sure that they are all correct (and look good).

docs/tutorials/Remark.adoc
j-lum: Yeah I've changed the commits to reflect that but the links are not updated yet. If there's nothing else wrong with the code we can request for @damithc to pull [this branch](https://github.com/j-lum/addressbook-level3/commits/tutorial-add-remark) in this sem's AB-3 and I'll update all the links to match. Can you help me do a quick once over on the commits on the branch as well?

damithc:
> This should go under a special type of documentation called "discussions". As much as possible we try to avoid adding text that does not explicitly push them towards the completion of the tutorial. We might want to cover this during the actual in-person tutorial?

When the tutorial says 'do x', it should also say why x is needed in order to move towards the goal (but not why we chose to do x instead of y). This is because the tutorial aims to _teach_ how to achieve the goal, not like _how-to_ documents that just list the steps. The _discussion_ type documentation covers things like 'x vs y'.

Best not to leave anything to be covered by the tutor. Some tutors will not have enough familiarity with the tutorial to do that. Besides, I'm not sure if we will be discussing these tutorials in the actual tutorial session.

damithc:
image

j-lum:
Bumping this up again since a student in the current semester ran into some formatting warnings from the usage of ``` as a code block fence.

damithc (changes requested):
Can port to Markdown? Remember not to wrap markdown lines.
For consistency with the other docs, use `1.` for all numbered items, `* ` for all bullets, and `_` for italics.

PR #26, opened 2019-07-30
Merge AB-4 "Migrate to PlantUML from PowerPoint"

j-lum:
* commit '5272d589ab491ea68fdf98d341c8ec6570663320':
  Migrate to PlantUML from PowerPoint


pyokagan:
@j-lum No need to use CIHR, it was designed at a time before we had this "merge AB-4 into AB-3 thing", so it's not equipped to deal with it. In fact, I'm still trying to work out how the workflow should be like.

We'll need to use a merge commit for this, so that next time we merge from AB-4 again Git would be able to use the merge as a suitable merge base commit. If we don't record this down as a merge commit in history, Git will keep flagging the same parts of the file as conflicting in the future.

btw, I think you can push directly to AB-3 master, unless you want me to take a look that the merge resolution is good before merging.

pyokagan:
> No need to use CIHR

In fact, we shouldn't use GitHub's "Merge pull request" button at all, since GitHub doesn't support fast-forward merges and we'll end up with *two* merge commits, which is not what we sensibly want :thinking:

damithc:
> In fact, we shouldn't use GitHub's "Merge pull request" button at all, since GitHub doesn't support fast-forward merges and we'll end up with _two_ merge commits, which is not what we sensibly want 🤔

@pyokagan did you mean merge with `--no-commit` option so that we get a chance to override the merge choices made by git?

pyokagan:
> @pyokagan did you mean merge with --no-commit option so that we get a chance to override the merge choices made by git?

You mean as opposed to using GitHub's "Merge Pull Request" button?

Right, that is one option. Generally, Git doesn't care what exact commands are used as long as we end up with the desired result :-)

j-lum:
@pyokagan @damithc I don't seem to have permission to push to AB3. 
[My master](https://github.com/j-lum/addressbook-level35/commits/master) seems to have preserved the merge properly though.

damithc:
> @pyokagan @damithc I don't seem to have permission to push to AB3.

Gave you write permissions just now.

pyokagan:
> [My master](https://github.com/j-lum/addressbook-level35/commits/master) seems to have preserved the merge properly though.

Heh, I see a lot of sneakiness in `LogicClassDiagram.puml` ;-)

I'll think about making the conflict resolution part of the normal AB-4 review workflow, so that these kind of changes can be reviewed as well. Might need to create some tooling to make that possible though, since GitHub provides zero support for it >:(

j-lum:
@pyokagan Removing CommandHistory caused the diagram to collapse into something completely unreadable.

Also is there some branch protection setting that I need to know about? My pushes are still failing for some reason.

pyokagan:
> Heh, I see a lot of sneakiness in LogicClassDiagram.puml ;-)

To further explain: I'm slightly worried about the impact of introducing too much "evilness" in a merge wrt future conflict resolutions, since such parts may conflict again in the future, but when devs search for a responsible commit in AB-3 that introduced the conflict (so that they, via reading the commit message, have the context to resolve it reasonably), they won't find any since it was done in a merge commit.

pyokagan:
> Removing CommandHistory caused the diagram to collapse into something completely unreadable.

Right, maybe a more suitable approach would be to do an obvious merge resolution (that causes the "collapse into something completely unreadable"), and then do a commit on top that fixes it, with a suitable commit message?

j-lum:
I was kind of worried about pushing something unreadable because I recalled something about not having a commit that leaves stuff in a broken state. Will make an exception for weird merges like this in the future.

pyokagan:
> Also is there some branch protection setting that I need to know about? My pushes are still failing for some reason.

Overzealous branch protection settings. I've fixed it. Can try again.

j-lum:
>
>
> > Also is there some branch protection setting that I need to know about? My pushes are still failing for some reason.
> 
> Overzealous branch protection settings. I've fixed it. Can try again.

Strange, I'm still getting 
```
remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: 2 of 2 required status checks are expected.
```

pyokagan:
@j-lum 

> ```
> remote: error: GH006: Protected branch update failed for refs/heads/master.
> remote: error: 2 of 2 required status checks are expected.
> ```

Alright, I've just removed branch protection entirely.

pyokagan:
> I was kind of worried about pushing something unreadable because I recalled something about not having a commit that leaves stuff in a broken state. 

Yeah, but in this case I'm thinking that arguably the damage is already done since we can't do a simple merge from AB-4 to AB-3 :-(

Anyway I don't think there is a clear cut answer to this, and I'll need to think more about it. For this, I'll let you handle as you see fit :-) :-)

j-lum:
Pushed successfully to master.

PR #27, opened 2019-08-05
Remove dependency on WebView

j-lum:
The `help` command is the only command that relies on WebView, a
dependency that weighs in at around 70MB, more than quadrupling the size
of the executable jar. As there are plans to customize the jars for each
individual student during the practical examination, this overhead can
lead to further problems (bandwidth in the lecture hall, storage space,
etc).

Let's remove the dependency on WebView by changing the `help` command to
output a URL to the user guide instead.

PR #28, opened 2019-08-29
Prevent HelpWindow from maximizing

j-lum:
Address a regression where subsequent `help` commands will spawn a
`HelpWindow` maximized, covering the screen.
Disable the maximize button completely as it serves no purpose.

### DIFF
--- a/src/main/java/seedu/address/ui/HelpWindow.java
+++ b/src/main/java/seedu/address/ui/HelpWindow.java
@@ -35,7 +35,6 @@ public class HelpWindow extends UiPart<Stage> {
     public HelpWindow(Stage root) {
         super(FXML, root);
         helpMessage.setText(HELP_MESSAGE);
-        root.sizeToScene();
     }
 
     /**
--- a/src/main/resources/view/HelpWindow.fxml
+++ b/src/main/resources/view/HelpWindow.fxml
@@ -8,9 +8,7 @@
 <?import javafx.scene.layout.HBox?>
 <?import javafx.stage.Stage?>
 
-<!-- TODO: set a more appropriate initial size -->
-
-<fx:root maximized="true" title="Help" type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/11" xmlns:fx="http://javafx.com/fxml/1">
+<fx:root resizable="false" title="Help" type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/11" xmlns:fx="http://javafx.com/fxml/1">
   <icons>
     <Image url="@/images/help_icon.png" />
   </icons>


src/main/java/seedu/address/ui/HelpWindow.java
```diff
@@ -36,6 +36,7 @@ public HelpWindow(Stage root) {
         super(FXML, root);
         helpMessage.setText(HELP_MESSAGE);
         root.sizeToScene();
+        root.setResizable(false);
```
pyokagan: This can be set in the FXML file so let's do it there so that all UI layout stuff is kept in one file.

src/main/resources/view/HelpWindow.fxml
```diff
@@ -10,7 +10,7 @@
 
 <!-- TODO: set a more appropriate initial size -->
 
-<fx:root maximized="true" title="Help" type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/11" xmlns:fx="http://javafx.com/fxml/1">
+<fx:root maximized="false" title="Help" type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/11" xmlns:fx="http://javafx.com/fxml/1">
```
pyokagan: The "maximized" attribute can be removed because "false" is the default. Gluon SceneBuilder would have removed it as well.

pyokagan (changes requested):
> ```
> Prevent HelpWindow from maximizing
> 
> Address a regression where subsequent `help` commands will spawn a
> `HelpWindow` maximized, covering the screen.
> Disable the maximize button completely as it serves no purpose.
> ```

If it's classified as a regression, then the commit which introduced the regression needs to be blamed. In this case,
ebbb7131ecdb83f2512b79068f952d37f69784b5.

src/main/java/seedu/address/ui/HelpWindow.java
```diff
@@ -36,6 +36,7 @@ public HelpWindow(Stage root) {
         super(FXML, root);
         helpMessage.setText(HELP_MESSAGE);
         root.sizeToScene();
```
pyokagan: btw, please check if this line is still needed. I'm guessing it was added to override the "maximize" behavior, but it may not actually be needed once we set `maximize="false"` since according to the docs at https://docs.oracle.com/javase/8/javafx/api/javafx/scene/Scene.html JavaFX is supposed to, by default, size scenes to the size of their content.

src/main/resources/view/HelpWindow.fxml
j-lum: Yeah somehow scene builder now absolutely refuses to open the FXML files. I might need to check if it's a version issue so we can peg a version of scene builder for students. I'll go check it out for sure.

pyokagan: fwiw scene builder works for me. I'm using the version 11 one.

j-lum: I'll go make a new PR that bumps JavaFX to version 11.0.1 before I go peg a version of Scene Builder I guess.

src/main/resources/view/HelpWindow.fxml
```diff
@@ -10,7 +10,7 @@
 
 <!-- TODO: set a more appropriate initial size -->
```
pyokagan: By the way, is this comment still relevant?

pyokagan (changes requested):
> ```
> Prevent HelpWindow from maximizing
> 
> Address a regression in PR #27 where subsequent `help` commands will
> ```

Use commit hashes instead or use a hyperlink, other Git tools usually don't recognize GitHub PR
references.

> ```
> spawn a `HelpWindow` maximized, covering the screen.
> Disable the maximize button completely as it serves no purpose.
> ```

Doesn't the removal of `sizeToScene()` need to be addressed here as
well?

src/main/resources/view/HelpWindow.fxml
j-lum: Kinda, we're still not setting a fixed size (mainly because I foresee students just replacing the link and having the label overflow).

pyokagan: > mainly because I foresee students just replacing the link and having the label overflow

Assuming that students do something like:
```diff
-    public static final String USERGUIDE_URL = "https://se-education.org/addressbook-level3/UserGuide.html";
+    public static final String USERGUIDE_URL = "https://my.new.website/addressbook-level3/UserGuide.html";
```

As long as the label contents are set before the window is first shown, the window will adapt its size to the size of the label, no?

pyokagan: > As long as the label contents are set before the window is first shown, the window will adapt its size to the size of the label, no?

Otherwise, the call to `root.sizeToScene()` *shouldn't* be removed?

pyokagan (approved):
Well, OK, I don't really have much time nowadays so I'll just leave some comments before I go:

> ```
> Prevent HelpWindow from maximizing
> 
> Address a regression in ebbb713 where subsequent `help` commands will
> spawn a `HelpWindow` maximized, covering the screen.
> ```

Previously (when the help window was still a web browser), it was designed to
be maximized when shown. It's not stated in the paragraph above, but I assume
that ebbb713 intended to make the window a fixed size non-resizable window, but
forgot to do so. It would be good if it was explicitly confirmed in the commit
message though.

> ```
> Disable the maximize button completely as it serves no purpose.
> ```

I guess even more importantly is that *resizing is completely disabled*, the
maximize button is just a side-effect. It's unclear from this paragraph whether
the capability of resizing "serves no purpose" as well. (Because if it does
serve some purpose, perhaps it might be better to remove the `maximized="true"`
but allow resizing).

> ```
> Refactor Java code pertaining to size and style to be FXML properties in
> line with existing convention.
> ```

This paragraph is a bit problematic, I might ask for changes if I was doing CS3282 or something :-)
Considering the diff is just:

```diff
diff --git a/src/main/java/seedu/address/ui/HelpWindow.java b/src/main/java/seedu/address/ui/HelpWindow.java
index 7a27ad09..9a665915 100644
--- a/src/main/java/seedu/address/ui/HelpWindow.java
+++ b/src/main/java/seedu/address/ui/HelpWindow.java
@@ -35,7 +35,6 @@
     public HelpWindow(Stage root) {
         super(FXML, root);
         helpMessage.setText(HELP_MESSAGE);
-        root.sizeToScene();
     }
 
     /**
diff --git a/src/main/resources/view/HelpWindow.fxml b/src/main/resources/view/HelpWindow.fxml
index a16c2852..a98aaa8f 100644
--- a/src/main/resources/view/HelpWindow.fxml
+++ b/src/main/resources/view/HelpWindow.fxml
@@ -10,7 +10,7 @@
 
 <!-- TODO: set a more appropriate initial size -->
 
-<fx:root maximized="true" title="Help" type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/11" xmlns:fx="http://javafx.com/fxml/1">
+<fx:root resizable="false" title="Help" type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/11" xmlns:fx="http://javafx.com/fxml/1">
   <icons>
     <Image url="@/images/help_icon.png" />
   </icons>
```

it's hard to see where the refactoring is. The removal of `maximized="true"`
and the addition of `resizable="false"` has already been addressed above, so
the only thing that remains is the removal of `root.sizeToScene()`, but it's
not being converted into FXML properties or anything. Puzzled.

src/main/resources/view/HelpWindow.fxml
j-lum: I was referring to a fixed size like in MainWindow where the minHeight and minWidth are defined. I do agree that we can completely do without and use the computed sizes though. I'll just remove the comment for now.

j-lum:
> This paragraph is a bit problematic, I might ask for changes if I was doing CS3282 or something :-)

I'm actually considering CS3282 next sem...

j-lum:
Shoving this through for now...

PR #41, opened 2020-05-25
Update checkstyle version to match Duke

j-lum:
Let's make the version of checkstyle consistent across projects.
Bump checkstyle to version 8.23.
Remove references to FileContentsHolder as it has been removed in v8.2.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ repositories {
 }
 
 checkstyle {
-    toolVersion = '8.1'
+    toolVersion = '8.29'
 }
 
 test {
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -29,12 +29,13 @@
     <property name="file" value="${config_loc}/suppressions.xml"/>
   </module>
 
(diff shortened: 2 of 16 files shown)


damithc:
@j-lum Let me know when ready for review

damithc:
@j-lum confirmed this config file works for IDEA as well? IIRC, that's where the version incompatibility caused problems.

j-lum:
Since there are a lot breaking changes to JavaDoc checks, I've updated them to match the latest version. There are some rules that don't have a direct 1-to-1 replacement but those don't seem to affect our code base much. (See for details: https://github.com/checkstyle/checkstyle/issues/7329) 

I've also added a check for `InvalidJavaDocPosition` which actually caught one such comment in one of our tests. 
When this gets merged I'll port over the checkstyle rules to Duke to ensure consistency. 

Also I've tested that the config file can be imported into the plugin and all inspections work.
image

damithc (changes requested):
I'm OK with the changes. This change is significant enough to warrant a proper commit message as per SE-EDU commit message standards. Can update the commit message?

PR #45, opened 2020-07-01
Display page title

### LINKED ISSUE #44: Site title does not follow the value in _config.yml
Although the title is given as `AB-3` in the `_config.yml`, the [generated website](https://se-education.org/addressbook-level3/) doesn't seem to reflect it.

j-lum:
Fixes #44 

The `<title>` tag's value is now set to the `title` attribute of each page. 
A live preview is available [here](https://j-lum.github.io/addressbook-level3/).

### DIFF
--- a/docs/_includes/head.html
+++ b/docs/_includes/head.html
@@ -7,4 +7,6 @@
 
   {%- include custom-head.html -%}
 
+  <title>{{page.title}}</title>
+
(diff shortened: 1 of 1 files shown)


damithc:
@j-lum Ready? If so, can propose a commit message? Is CIHR still not working?

PR #50, opened 2020-09-15
Change IDEA codestyle href

Persdre:
On the "Setting up and getting started" page of AB3 developer guide (https://nus-cs2103-ay2021s1.github.io/tp/SettingUp.html), the href for "[se-edu/guides] IDEA: Configuring the JDK" is pointed towards https://se-education.org/guides/tutorials/checkstyle.html, instead of https://se-education.org/guides/tutorials/intellijCodeStyle.html


damithc:
Thanks for the fix @Persdre 💯

PR #63, opened 2020-12-02
Remove unnecessary comments

CodyChew:
Irrelevant comments found in test cases may cause confusion to future developers.

Let's remove `UndoCommand` and `RedoCommand` in comments

### DIFF
--- a/src/test/java/seedu/address/logic/commands/DeleteCommandTest.java
+++ b/src/test/java/seedu/address/logic/commands/DeleteCommandTest.java
@@ -19,7 +19,7 @@ import seedu.address.model.UserPrefs;
 import seedu.address.model.person.Person;
 
 /**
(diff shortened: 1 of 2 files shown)


damithc:
Thanks for the PR @CodyChew 
We'll merge it in due course.

damithc:
@CodyChew do you mind submitting an update as noted in https://github.com/se-edu/addressbook-level3/pull/63#issuecomment-737188116 ?
Furthermore, remember to follow the commit message format as per [this guide](https://se-education.org/guides/conventions/git.html), both for the subject and the body.

damithc:
Thanks for the fix @CodyChew 
Note that I modified the commit message a bit.

CodyChew:
> Note that I modified the commit message a bit.

Thanks @damithc, I will learn from it

PR #64, opened 2020-12-04
Fix notation mistake in Model class diagram

### LINKED ISSUE #56: DG: wrong multiplicity in the Model class diagram
https://github.com/nus-cs2103-AY2021S1/forum/issues/398

wltan:
Fixes #56. As pointed out in the [forum post](https://github.com/nus-cs2103-AY2021S1/forum/issues/398),

> The multiplicity should be * and the association role `filtered list` should appear near the end of the arrow.

The updated diagram now looks like this, fixing both problems:
image

Note that the multiplicity and association role are now part of the same "label". To my knowledge, this is the only way to put both multiplicity and association role at the same end of the arrow in PlantUML. [This StackOverflow answer](https://stackoverflow.com/questions/60226784/in-plantuml-how-do-i-specify-role-names-at-either-end-of-an-association-between) also suggests such an implementation.


damithc:
@j-lum can review this as you are the PlantUML expert?

damithc:
@se-edu/tech-team-level1 any reviews for this PR?

ernestlim8:
I pulled the fork of the PR by OP and I did not manage to replicate the same diagram when I exported the .puml file as PNG.

image

I increased the length of the arrow, which automatically caused the association label to adjust into the desired position.

I edited line 55 of ModelClassDiagram.puml
`ModelManager ----->" ~* filtered list" Person`

Attached below is the result of this code.
image

The branch name of OP does not meet the se-edu conventions as it does not contain the issue number. The branch might be better renamed as 56-fix-notation-model-diagram.

damithc:
@wltan are you able to resume this PR? If can, I want to merge it by this weekend.

wltan:
@ernestlim8 @damithc

I have made the changes highlighted in https://github.com/se-edu/addressbook-level3/pull/64#issuecomment-759245128 and re-exported the image file. If the diagram is still not identical, then I must have messed something up in my exporting workflow.

As for the branch name, I am not aware of any way to rename my branch without having to create a new PR, which is likely undesirable at this stage.

damithc:
> I have made the changes highlighted in [#64 (comment)](https://github.com/se-edu/addressbook-level3/pull/64#issuecomment-759245128) and re-exported the image file. If the diagram is still not identical, then I must have messed something up in my exporting workflow.

Thanks @wltan @ernestlim8 can check again?

> As for the branch name, I am not aware of any way to rename my branch without having to create a new PR, which is likely undesirable at this stage.

Yes, let's leave it be.

ernestlim8:
@wltan I've pulled your branch and the exported image is correct on my end.

damithc:
Thanks for the fix @wltan and thanks @ernestlim8 for the inputs.
I added some details to the commit message when merging.

PR #68, opened 2020-12-05
Update checking for Duplicates in EditCommand

### LINKED ISSUE #57: Allows editing a contact to become a duplicate of another
Reported in a student project but it seems the bug originates in AB3:
https://github.com/nus-cs2103-AY2021S1/pe-dev-response/issues/5266
https://github.com/nus-cs2103-AY2021S1/pe-dev-response/issues/5288

Steps to reproduce:
1. edit 1 n/Bernice Yu
2. edit 1 p/99272758
3. edit 1 e/berniceyu@example.com
4. edit 1 t/colleagues t/friends
5. edit 1 a/Blk 30 Lorong 3 Serangoon Gardens, #07-18

**Result:** 
The app allowed duplicate contacts
image

Also, when we closed and reopen the app, the app started empty (because the data file had been corrupted by duplicate contacts)

image

CodyChew:
fixes #57

Let's 
* use a separate Model with only the relevant Persons to check for Duplicates.
* include test cases for EditCommand

The check for duplicate Persons comes from the isSamePerson method in Person class which checks for the same name and having at least one other identity field(Phone or Email) being the same. To reuse the checking mechanism from the Model, a simple way would be to remove the personToEdit from a model and have a separate check.

### DIFF
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -19,6 +19,7 @@ import seedu.address.commons.core.index.Index;
 import seedu.address.commons.util.CollectionUtil;
 import seedu.address.logic.commands.exceptions.CommandException;
 import seedu.address.model.Model;
+import seedu.address.model.ModelManager;
 import seedu.address.model.person.Address;
 import seedu.address.model.person.Email;
 import seedu.address.model.person.Name;
@@ -77,7 +78,9 @@ public class EditCommand extends Command {
         Person personToEdit = lastShownList.get(index.getZeroBased());
         Person editedPerson = createEditedPerson(personToEdit, editPersonDescriptor);
 
-        if (!personToEdit.isSamePerson(editedPerson) && model.hasPerson(editedPerson)) {
+        Model modelWithoutPersonToEdit = new ModelManager(model.getAddressBook(), model.getUserPrefs());
+        modelWithoutPersonToEdit.deletePerson(personToEdit);
+        if (modelWithoutPersonToEdit.hasPerson(editedPerson)) {
             throw new CommandException(MESSAGE_DUPLICATE_PERSON);
         }
 
--- a/src/test/java/seedu/address/logic/commands/EditCommandTest.java
+++ b/src/test/java/seedu/address/logic/commands/EditCommandTest.java
@@ -4,7 +4,10 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static seedu.address.logic.commands.CommandTestUtil.DESC_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.DESC_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_EMAIL_AMY;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_NAME_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_NAME_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_PHONE_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_PHONE_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_HUSBAND;
 import static seedu.address.logic.commands.CommandTestUtil.assertCommandFailure;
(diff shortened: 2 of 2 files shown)


damithc:
@CodyChew thanks for the fix.
Did you also check the info given in https://github.com/se-edu/addressbook-level3/issues/57#issuecomment-739196585 ?

CodyChew:
@damithc I just checked it out. I thought of it as the checking at the Command being problematic rather than the equality check in `Person`. I extracted out the checking to only the *relevant* model and do not have to deal with the transitive property of `isSamePerson`. Since the checking is only done at one place, it's less prone to bugs and `isSamePerson` can still be customised.

This check is similar to `addCommand` where it checks the person to add against the model *before* adding it, however it requires the creation of an additional model for the check within `editCommand`.

damithc:
@fzdy1914 can take a look at this PR?

fzdy1914 (commented):
I am not so sure adding another model is elegant enough. Basically, it requires duplicating the whole data, which is usually unacceptable in a larger project.

There are two ways possible, one is to add another method for checking duplicate person in Addressbook. Second is to delete the person first, do the checking, restore the person if some error happens. 

@damithc Prof, you may decide which solution is better.

#69 is using the method 1.

damithc:
> I am not so sure adding another model is elegant enough. Basically, it requires duplicating the whole data, which is usually unacceptable in a larger project.
> 
> There are two ways possible, one is to add another method for checking duplicate person in Addressbook. Second is to delete the person first, do the checking, restore the person if some error happens.
> 
> @damithc Prof, you may decide which solution is better.
> 
> #69 is using the method 1.

Yes, duplicating the entire model seems excessive. We can try one of the other two methods if they are not a lot more complicated than the one used in this PR. But I don't mind an inelegant/inefficient solution if the implementation is a lot simpler and the performance cost is not prohibitive.

@CodyChew what do you think about the other two solutions?

CodyChew:
> There are two ways possible, one is to add another method for checking duplicate person in Addressbook. Second is to delete the person first, do the checking, restore the person if some error happens.

I prefer the second approach since we would still be able to use the same `isSamePerson` method without introducing different/additional methods for checking. The check will be simpler and it will be less prone to bugs since the check for `addCommand` and `editCommand` will be *similar*.

damithc:
> I prefer the second approach since we would still be able to use the same `isSamePerson` method without introducing different/additional methods for checking. The check will be simpler and it will be less prone to bugs since the check for `addCommand` and `editCommand` will be _similar_.

Noted. Perhaps you can submit a separate PR using that approach so that we can compare all three?

CodyChew:
I have made a PR using the second approach, one main issue I came across was that adding and deleting a `Person` using the current commands would result in different internal lists due to ordering. I have tried to work around it by extending the functionality of the model, please refer to #74 .

damithc:
> @CodyChew thanks for the fix.
> Did you also check the info given in [#57 (comment)](https://github.com/se-edu/addressbook-level3/issues/57#issuecomment-739196585) ?

I had another look at this comment. Seems the recommendation was to simplify the `isSamePerson()`. Shouldn't that be simpler than any of the alternatives considered here?

CodyChew:
Yes, simplifying the method would resolve the issue too. However if we want to keep the current definition of `isSamePerson()` and allow students to customize their own method behavior, changing the way we check in `editCommand` would make it  independent of the nature of `isSamePerson()` and reduce such bugs.

Definition of `isSamePerson()`:
`* Returns true if both persons of the same name have at least one other identity field that is the same.`
`* This defines a weaker notion of equality between two persons.`

damithc:
> Yes, simplifying the method would resolve the issue too. However if we want to keep the current definition of `isSamePerson()` and allow students to customize their own method behavior, changing the way we check in `editCommand` would make it independent of the nature of `isSamePerson()` and reduce such bugs.
> 
> Definition of `isSamePerson()`:
> `* Returns true if both persons of the same name have at least one other identity field that is the same.`
> `* This defines a weaker notion of equality between two persons.`

I agree that improving the check makes the code more flexible. However, if we use a more sophisticated equality check that students might not even need in the end, it can add an unnecessary bump to the learning curve, especially for weaker students. Therefore, I'm inclined to go with the simpler solution of simplifying the `isSamePerson()`. Those who need a more complex equality can add one at their own expense.  We can add a `TODO` comment in the code to mention that a more sophisticated equality check can be considered for future versions. It will be a good exercise for students to figure out the implications of such a non-transitive equality relation, uncover this potential bug, and fix it. What do you think?

CodyChew:
I agree that if we make `isSamePerson()` to only check the name, it will be easier for them to understand and expand on the method. They would then have to change the guard condition themselves if they encounter this bug should they make `isSamePerson()` more complex. 

 Should I make another PR for this solution? Some of the test cases and  *default Person* in `PersonBuilder` would need to be changed, now that there cannot be Persons of the same name.

damithc:
> Should I make another PR for this solution? Some of the test cases and _default Person_ in `PersonBuilder` would need to be changed, now that there cannot be Persons of the same name.

Yes, go ahead @CodyChew 
On a related note, whichever PR we merge, we should give the other 3 as references in the commit message.

damithc:
Closing in favor of #75

PR #69, opened 2020-12-05
Bug fix edit command

itssodium:
Fixed the Edit Command bug that allowed duplicate people to be added. 

The bug resided in the following portion

`if (!personToEdit.isSamePerson(editedPerson) && model.hasPerson(editedPerson)) {
            throw new CommandException(MESSAGE_DUPLICATE_PERSON);
        }`

The `personToEdit.isSamePerson(editedPerson)` gives `true` when either the email or phone number is different. Therefore, when editing a person with an already existing name and changing the phone number to an already existing phone number this results in true, but since we are editing it we may need to check all the attributes.

Due to the above change `model.hasPerson(editedPerson)` may give true when an editPerson refers to the same original person, due to the reason stated above. Therefore, I have changed it such that you only compare against InternalList with the person at index stated by person removed.

### DIFF
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -77,7 +77,7 @@ public class EditCommand extends Command {
         Person personToEdit = lastShownList.get(index.getZeroBased());
         Person editedPerson = createEditedPerson(personToEdit, editPersonDescriptor);
 
-        if (!personToEdit.equals(editedPerson) && model.hasPerson(editedPerson)) {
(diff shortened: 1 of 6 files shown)


damithc:
@itssodium thanks for the PR
Looks like there is an earlier PR for the same issue #68 
I'm putting this on hold for now

damithc:
@itssodium perhaps you can compare your approach to that of #68 ? For example, you can add your questions/comments to that PR -- that is also a good way of contributing (contributions doesn't always have to your own PR)

damithc:
Closing in favor of #75

PR #70, opened 2020-12-06
Edit absolute links in Developer Guide to be relative links

### LINKED ISSUE #51: DG: remove absolute links to AB3 repo
DG contains absolute links e.g., `https://github.com/se-edu/addressbook-level3/tree/master/src/main/java/seedu/address/Main.java`

Perhaps we can make the relative links so that they don't need changing after forking?

mehak24k:
Fixes #51

### DIFF
--- a/docs/DeveloperGuide.md
+++ b/docs/DeveloperGuide.md
@@ -23,11 +23,11 @@ The ***Architecture Diagram*** given above explains the high-level design of the
 
 <div markdown="span" class="alert alert-primary">
 
-:bulb: **Tip:** The `.puml` files used to create diagrams in this document can be found in the [diagrams](https://github.com/se-edu/addressbook-level3/tree/master/docs/diagrams/) folder. Refer to the [_PlantUML Tutorial_ at se-edu/guides](https://se-education.org/guides/tutorials/plantUml.html) to learn how to create and edit diagrams.
+:bulb: **Tip:** The `.puml` files used to create diagrams in this document can be found in the [diagrams](././diagrams) folder. Refer to the [_PlantUML Tutorial_ at se-edu/guides](https://se-education.org/guides/tutorials/plantUml.html) to learn how to create and edit diagrams.
 
 </div>
 
-**`Main`** has two classes called [`Main`](https://github.com/se-edu/addressbook-level3/tree/master/src/main/java/seedu/address/Main.java) and [`MainApp`](https://github.com/se-edu/addressbook-level3/tree/master/src/main/java/seedu/address/MainApp.java). It is responsible for,
+**`Main`** has two classes called [`Main`](././../src/main/java/seedu/address/Main.java) and [`MainApp`](././../src/main/java/seedu/address/MainApp.java). It is responsible for,
 * At app launch: Initializes the components in the correct sequence, and connects them up with each other.
 * At shut down: Shuts down the components and invokes cleanup methods where necessary.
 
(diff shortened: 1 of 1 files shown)


damithc:
Thanks for the fix @mehak24k 
Can you submit an update as per the bot comment above and also provide a full commit message as per https://se-education.org/guides/conventions/git.html ?

mehak24k:
> Thanks for the fix @mehak24k
> Can you submit an update as per the bot comment above and also provide a full commit message as per https://se-education.org/guides/conventions/git.html ?

Thank you for the feedback, I have refined the commit message and submitted an update with the bot.

siangernlow:
While the proposed changes are valid, the links break on the published GitHub Pages site due to the different domain name. As such, the links would have to be changed anyway once the repository is forked.

A potential solution would be to use link references as specified [here](https://kramdown.gettalong.org/quickref.html#links-and-images), but this is also unviable as the concatenation of links is not supported. That means that the references would have to be absolute links anyway and would not solve the issue.

damithc:
> A potential solution would be to use link references as specified [here](https://kramdown.gettalong.org/quickref.html#links-and-images), but this is also unviable as the concatenation of links is not supported. That means that the references would have to be absolute links anyway and would not solve the issue.

@siangernlow You might be right that they have to remain absolute links. If no one else can come up with a solution in the next few days, we can give up on this issue and close this PR.

siangernlow:
@damithc I have found a potential solution that could be considered. 

We can define a variable in the front matter of `DeveloperGuide.md` which stores the repo path. Subsequently, the absolute links would then reference this variable. As a result, when the project is forked, only the value stored by the variable would have to be changed. I've included some images below to illustrate this.

image
_Creating a variable `repo` to store the path to be referenced_

image
_Referencing the variable_

Referring to above, when the project is forked, only the path stored in `repo` would have to change.

damithc:
> @damithc I have found a potential solution that could be considered.
> 
> We can define a variable in the front matter of `DeveloperGuide.md` which stores the repo path. Subsequently, the absolute links would then reference this variable. As a result, when the project is forked, only the value stored by the variable would have to be changed. I've included some images below to illustrate this.

If this is a site-wide setting, should we use the method in [here](https://se-education.org/guides/tutorials/jekyll.html#site-wide-settings) instead?

siangernlow:
> If this is a site-wide setting, should we use the method in [here](https://se-education.org/guides/tutorials/jekyll.html#site-wide-settings) instead?

Yes, we could use that method. In the _config.yml file, there is already a `repository` variable which could be used for the purpose.

damithc:
> Yes, we could use that method. In the _config.yml file, there is already a `repository` variable which could be used for the purpose.

Ah, good. Do submit a PR if you manage to get it working.

siangernlow:
Noted!

siangernlow:
It appears that GitHub's preview does not parse Liquid tags which are required to reference the variables. This causes the preview on GitHub to have broken links.  So I guess in the end absolute links are the way to go.

damithc:
> It appears that GitHub's preview does not parse Liquid tags which are required to reference the variables. This causes the preview on GitHub to have broken links. So I guess in the end absolute links are the way to go.

I see. You mentioned that the variable is already present in the config file. How is it used currently? Doesn't the same problem apply to those usages?

siangernlow:
I believe that the variable is being used by Jekyll in HTML files (`header.html`) to generate the published site. In this case, Jekyll is able to parse the Liquid tags, so there are no problems for that. This is also the reason why the proposed solution works on the published site, but not in the preview on GitHub.

damithc:
> I believe that the variable is being used by Jekyll in HTML files (`header.html`) to generate the published site. In this case, Jekyll is able to parse the Liquid tags, so there are no problems for that. This is also the reason why the proposed solution works on the published site, but not in the preview on GitHub.

Got it. I guess no choice but to leave the absolute links as they are. Thanks for the investigation @siangernlow

PR #72, opened 2020-12-07
Update UG to address issues #58 #62

### LINKED ISSUE #58: Document the behavior of accepting a command parameter multiple times
Currently, if a command contains the same parameter e.g., `p/` multiple times, the last value will be taken. Rationale: this allows the user to rectify erroneous parameter values on the fly, without having to backspace.

Suggested: 
* Document this behavior as it can be misinterpreted as a bug
* When this happens, the command output can contain a friendly warning that multiple values were found in the command, in case the repetition was unintentional

### LINKED ISSUE #62: UG: mention security setting in case of using Mac
Double-clicking the jar file might not work in Mac computers.

Suggested: mention how to overcome it in the UG

Zhang-Yunjie:
Closes #62 
Closes #58

### DIFF
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -23,7 +23,6 @@ AddressBook Level 3 (AB3) is a **desktop app for managing contacts, optimized fo
   * Open a command window in that folder
   * Run the command java -jar {filename}.jar <br>e.g., java -jar AddressBook3.jar (i.e., run the command in the same folder as the jar file)
 
-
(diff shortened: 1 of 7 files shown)


damithc:
Thanks for the fix @Zhang-Yunjie 
Let's do two separate PRs as these are two different fixes. Easier to manage and faster to merge that way.
Also, can you submit an update as per the bot comment above and also provide a full commit message as per https://se-education.org/guides/conventions/git.html ?

damithc:
Closing due to inactivity

PR #74, opened 2020-12-16
Update checking for Duplicates in EditCommand (Follow up from #68)

CodyChew:
Follow up from #68 

Let's
* remove the target `personToEdit` from the model
* add the `Person` back to its original position in the `AddressBook` after the check
* extend the functionality of the Model to add a `Person` at a specified `index`, it should also not update the filtered person list

The `UniquePersonList` in `AddressBook` currently does not support adding a Person at a specified position, however, it is needed for the check as we have to add back the deleted `personToEdit` to restore the exact internal list.

The overloaded `addPerson(int index, Person person)` should not update the filtered list as it will be used in cases where there are duplicates and an exception is returned. The filtered list do not need to be updated here.


damithc:
Closing in favor of #75

PR #75, opened 2020-12-18
Update isSamePerson() equality for Person class (fix #57)

### LINKED ISSUE #57: Allows editing a contact to become a duplicate of another

CodyChew:
Fixes #57  

Let's
* change isSamePerson() to return true for Persons with the same name
* change existing test cases to accommodate for the new behaviour

Refer to #68, #74, #69 for other alternatives

### DIFF
--- a/src/test/java/seedu/address/model/person/PersonTest.java
+++ b/src/test/java/seedu/address/model/person/PersonTest.java
@@ -31,27 +31,23 @@ public class PersonTest {
         // null -> returns false
         assertFalse(ALICE.isSamePerson(null));
 
-        // different phone and email -> returns false
-        Person editedAlice = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).withEmail(VALID_EMAIL_BOB).build();
-        assertFalse(ALICE.isSamePerson(editedAlice));
+        // same name, all other attributes different -> returns true
+        Person editedAlice = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).withEmail(VALID_EMAIL_BOB)
+                .withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_HUSBAND).build();
+        assertTrue(ALICE.isSamePerson(editedAlice));
 
-        // different name -> returns false
+        // different name, all other attributes same -> returns false
         editedAlice = new PersonBuilder(ALICE).withName(VALID_NAME_BOB).build();
         assertFalse(ALICE.isSamePerson(editedAlice));
 
-        // same name, same phone, different attributes -> returns true
-        editedAlice = new PersonBuilder(ALICE).withEmail(VALID_EMAIL_BOB).withAddress(VALID_ADDRESS_BOB)
-                .withTags(VALID_TAG_HUSBAND).build();
-        assertTrue(ALICE.isSamePerson(editedAlice));
+        // name differs in case, all other attributes same -> returns false
+        Person editedBob = new PersonBuilder(BOB).withName(VALID_NAME_BOB.toLowerCase()).build();
+        assertFalse(BOB.isSamePerson(editedBob));
 
-        // same name, same email, different attributes -> returns true
-        editedAlice = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).withAddress(VALID_ADDRESS_BOB)
-                .withTags(VALID_TAG_HUSBAND).build();
-        assertTrue(ALICE.isSamePerson(editedAlice));
-
-        // same name, same phone, same email, different attributes -> returns true
-        editedAlice = new PersonBuilder(ALICE).withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_HUSBAND).build();
-        assertTrue(ALICE.isSamePerson(editedAlice));
+        // name has trailing spaces, all other attributes same -> returns false
+        String nameWithTrailingSpaces = VALID_NAME_BOB + " ";
+        editedBob = new PersonBuilder(BOB).withName(nameWithTrailingSpaces).build();
+        assertFalse(BOB.isSamePerson(editedBob));
     }
 
     @Test
--- a/src/test/java/seedu/address/testutil/PersonBuilder.java
+++ b/src/test/java/seedu/address/testutil/PersonBuilder.java
@@ -16,9 +16,9 @@ import seedu.address.model.util.SampleDataUtil;
  */
 public class PersonBuilder {
 
-    public static final String DEFAULT_NAME = "Alice Pauline";
+    public static final String DEFAULT_NAME = "Amy Bee";
     public static final String DEFAULT_PHONE = "85355255";
-    public static final String DEFAULT_EMAIL = "alice@gmail.com";
+    public static final String DEFAULT_EMAIL = "amy@gmail.com";
     public static final String DEFAULT_ADDRESS = "123, Jurong West Ave 6, #08-111";
 
     private Name name;
--- a/src/main/java/seedu/address/model/person/Person.java
+++ b/src/main/java/seedu/address/model/person/Person.java
@@ -61,7 +61,7 @@ public class Person {
     }
 
     /**
-     * Returns true if both persons of the same name have at least one other identity field that is the same.
+     * Returns true if both persons have the same name.
      * This defines a weaker notion of equality between two persons.
      */
     public boolean isSamePerson(Person otherPerson) {
@@ -70,8 +70,7 @@ public class Person {
         }
 
         return otherPerson != null
-                && otherPerson.getName().equals(getName())
-                && (otherPerson.getPhone().equals(getPhone()) || otherPerson.getEmail().equals(getEmail()));
+                && otherPerson.getName().equals(getName());
     }
 
     /**


src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -119,6 +120,27 @@ public void execute_duplicatePersonFilteredList_failure() {
         assertCommandFailure(editCommand, model, EditCommand.MESSAGE_DUPLICATE_PERSON);
     }
 
+    @Test
+    public void execute_duplicateFieldsUnfilteredList_nameMustBeDifferent() {
```
damithc: Quick comment: This three-part test naming format has a convention for what each part should be.

CodyChew: I had troubles figuring out the name for the tests. I was testing the execute method with two edits to represent the bug, first editing a person with similar fields(success), and then editing to a similar name(failure). Would it be better if it was `execute_differentNameDuplicateFieldsUnfilteredList_success()`, but it seemed a bit long. Or maybe this test case isn't necessary since we do not need to test on the non-transitive nature anymore.

damithc: If you are testing both success and failure outcomes, you can omit the third part of the name.
Yes, this test case might be redundant.

CodyChew: Thanks for the tip, I have removed the test case and updated this commit.

src/test/java/seedu/address/testutil/PersonBuilder.java
```diff
@@ -16,9 +16,9 @@
  */
 public class PersonBuilder {
 
-    public static final String DEFAULT_NAME = "Alice Pauline";
+    public static final String DEFAULT_NAME = "Amy Bee";
     public static final String DEFAULT_PHONE = "85355255";
-    public static final String DEFAULT_EMAIL = "alice@gmail.com";
+    public static final String DEFAULT_EMAIL = "amy@gmail.com";
     public static final String DEFAULT_ADDRESS = "123, Jurong West Ave 6, #08-111";
```
damithc, line 22: Sorry @CodyChew, it took me a while to get back to this PR.
The PR seems much simpler compared to the other two alternatives.
However, it's not immediately apparent why the default values here need to change. I assume these constants are used in other test cases as well?

CodyChew: Yes @damithc, it is only failing in one test case in `AddCommandIntegrationTest`
```
public void execute_newPerson_success() {
        Person validPerson = new PersonBuilder().build();

        Model expectedModel = new ModelManager(model.getAddressBook(), new UserPrefs());
        expectedModel.addPerson(validPerson);

        assertCommandSuccess(new AddCommand(validPerson), model,
                String.format(AddCommand.MESSAGE_SUCCESS, validPerson), expectedModel);
}
```
This is because the `validPerson` uses the default name in `PersonBuilder.java` which is the same as the first person in the model which uses  `TypicalPersons.java`. This test case will work so long as the default name is not any name already in `getTypicalPersons()`

For reference:
```
public static List<Person> getTypicalPersons() {
        return new ArrayList<>(Arrays.asList(ALICE, BENSON, CARL, DANIEL, ELLE, FIONA, GEORGE));
}
```

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -31,9 +31,9 @@ public void isSamePerson() {
         // null -> returns false
         assertFalse(ALICE.isSamePerson(null));
```
damithc, line 33: Given that email no longer has a special significance in equality check, I think some of the test cases below are redundant now. Perhaps enough to test these two only?

* Same name, everything else is different -> true
* Different name, everything else is same -> false

What do you think?

damithc (changes requested):
Good work. Let's go with this alternative, but need to polish it up a bit more.

I think the commit message should be written along the following lines:
* The person equality check returns true if the name or the email matches. 
* The OR condition complicates the equality check done when editing a person. {describe the bug it causes} 
* Alternative 1:  Improve the way the equality check is done in the edit command to prevent the above bug. {Quote the two PRs that are variants of this approach.}
* Alternative 2: Simplify the equality check by removing the OR condition. This is simpler but possibly too strict for the address book domain i.e., increases the risk of duplicate entries for the same person creeping into the address book
* We decided to go with the alternative 2 because,
  * It is important to keep the code base simple, to prevent the initial learning curve of students at the start of their project
  * Students can introduce a more sophisticated equality check that matches their product domain later.

src/test/java/seedu/address/model/person/PersonTest.java
CodyChew: i agree, the following 3 would be redundant
* same name, same phone, different attributes -> returns true
* same name, same email, different attributes -> returns true
* same name, same phone, same email, different attributes -> returns true

I will remove these tests and update the commit accordingly

damithc:
@CodyChew ready for review? If so, pls submit a new version using the bot.

CodyChew:
> @CodyChew ready for review? If so, pls submit a new version using the bot.

Sorry, didn't see v2 and submitted it twice

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -31,27 +31,14 @@ public void isSamePerson() {
...
 
-        // different phone and email -> returns false
-        Person editedAlice = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).withEmail(VALID_EMAIL_BOB).build();
-        assertFalse(ALICE.isSamePerson(editedAlice));
+        // same name, different attributes -> returns true
```
damithc: nit, to be absolutely clear:
`//same name, all other attributes different -> returns true`

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -31,27 +31,14 @@ public void isSamePerson() {
...
+        Person editedAlice = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).withEmail(VALID_EMAIL_BOB)
+                .withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_HUSBAND).build();
+        assertTrue(ALICE.isSamePerson(editedAlice));
 
         // different name -> returns false
```
damithc: Similarly, `// different name -> all other attributes same -> returns false`

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -31,27 +31,14 @@ public void isSamePerson() {
...
         editedAlice = new PersonBuilder(ALICE).withName(VALID_NAME_BOB).build();
         assertFalse(ALICE.isSamePerson(editedAlice));
```
damithc, line 41: We can also consider these cases:
* name differs in case only
* name has leading/trailing spaces

Without the above, a future developer can add a `trim()` or a `toLowerCase()` to the logic without any changes to the test cases.

damithc (commented):
I think we are getting there @CodyChew. Some nits added. In case you haven't notice already, a PR to SE-EDU repos can take many iterations :-p

damithc:
@Zhiyuan-Amos do you have time to take a look at this one?

src/test/java/seedu/address/model/person/PersonTest.java
CodyChew: > Without the above, a future developer can add a trim() or a toLowerCase() to the logic without any changes to the test cases.

Do you mean if the developer directly modifies the `isSamePerson()` method with something like `toLowerCase()`? Yes, i think it's good to add a *name differs in case only test* to ensure case-sensitivity. But if I'm not wrong the leading and trailing spaces would already be handled by `ArgumentTokenizer.java` when it parses the user input.

`ArgumentTokenizer.java` documentation reference:
```
2. Leading and trailing whitespaces of an argument value will be discarded.<br>
```

Zhiyuan-Amos: > leading and trailing spaces would already be handled by ArgumentTokenizer.java

You are right, but `PersonTest` is a unit test; `isSamePerson()` should be tested independently of other components of the application, such as the `ArgumentTokenizer`. Therefore, `isSamePerson()` should not assume that there aren't leading and trailing spaces.

src/test/java/seedu/address/model/person/PersonTest.java
Zhiyuan-Amos: For consistency with the earlier comment https://github.com/se-edu/addressbook-level3/pull/75#discussion_r550386291, I think it should be this instead? :P

// different name ~~->~~**,** all other attributes same -> returns false

Zhiyuan-Amos:
Hi @CodyChew, thanks for your PR! Just 1 more comment apart from the comments I left in earlier threads: 

We [link our PRs to issue(s) using keywords like `Fixes #X`](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) in the PR *description*, so that the issue will be automatically closed when the PR is merged. So in this case, the `Fix #57` should not be at the PR *title*, but at the PR *description*.

Great job otherwise! 👍

src/test/java/seedu/address/testutil/PersonBuilder.java
Zhiyuan-Amos: I think it's helpful to explain in the commit message why this change was made, so that future readers of the commit can know why this change was made without having to refer to this thread on GitHub. It can be something like:

```
Also, let's change the default values of.... This is because AddCommandIntegrationTest...
```

Zhiyuan-Amos (changes requested):

src/test/java/seedu/address/model/person/PersonTest.java
damithc: Good points. Instead, should be add asserts inside the Person class to verify that fields don't contain any leading/trailing spaces? It can be a separate PR though.

CodyChew: Ahh I see @Zhiyuan-Amos, I have added the two test cases for `isSamePerson()`. I am not sure if it's correct, but I have decided to manipulate/append the Strings in the test cases rather than add another variable to `CommandTestUtil.java` for same name with different case/trailing spaces.
For the leading/trailing spaces test, I realised that I cannot test for leading spaces since the first character of a `Name` cannot start with a whitespace.

src/test/java/seedu/address/testutil/PersonBuilder.java
CodyChew: Should I squash the commits into a single commit? and revise the commit message

damithc:
@CodyChew yes, I think one commit is enough for this change as tests and code changes go together. Please get the bot to post a new version when ready.

src/test/java/seedu/address/testutil/PersonBuilder.java
Zhiyuan-Amos: Sorry, missed this comment. Currently this change is in the same commit as the other changes to isSamePerson() right? They can remain in the same commit, with the modification to the commit message as what I typed above.

damithc (changes requested):
@CodyChew thanks for the update. I couldn't follow this explanation in the commit message though.

>Also, let's change the default values of PersonBuilder utility class.
This is because AddCommandIntegrationTest uses the default name in
PersonBuilder and expects it to build a valid person. In this case, the
default name cannot be a name in a model which uses TypicalPersons.

Can improve it a bit more?

Also, wait for @Zhiyuan-Amos 's review before updating.

Zhiyuan-Amos (commented):
@CodyChew good work! I'm good with this version, just elaborate on the commit message as mentioned by Prof Damith.

This repo has strict requirements regarding commit messages, and it's actually helpful in the workplace when you explain code changes to higher-ups who often don't have the context as you do. :) no doubt this change is trivial, but it's a form of training I guess.

If you're unclear, you can suggest a commit message in a comment and tag me first, rather than having to git rebase to make changes and submit another iteration. :)

damithc:
@CodyChew let's wrap this up soon, before the semester starts. We are almost there.

CodyChew:
> @CodyChew good work! I'm good with this version, just elaborate on the commit message as mentioned by Prof Damith.
> 
> This repo has strict requirements regarding commit messages, and it's actually helpful in the workplace when you explain code changes to higher-ups who often don't have the context as you do. :) no doubt this change is trivial, but it's a form of training I guess.
> 
> If you're unclear, you can suggest a commit message in a comment and tag me first, rather than having to git rebase to make changes and submit another iteration. :)

Thanks for the advise @Zhiyuan-Amos. would this be okay to explain the changes to default values? The rest of the commit message will be the same.
```
Also, let's change the default name in PersonBuilder utility class.
This is because AddCommandIntegrationTest uses the default name to 
build a valid person and then adds it into an existing model. In this case, 
the existing model uses getTypicalPersons() which already has the default 
name. Thus we need to change the default name to a unique name that 
is not inside getTypicalPersons(). Also, the default email is changed for 
consistency in naming the default values, it does not affect current test cases.
```

Zhiyuan-Amos:
@CodyChew Your commit message looks good! Just a thought on structuring it even better. We can structure this segment as per how we usually structure commit messages:

```
<Current Situation>

<Problem>

<Solution>
```

So, I think the following segment:

> This is because AddCommandIntegrationTest uses the default name to 
build a valid person and then adds it into an existing model. In this case, 
the existing model uses getTypicalPersons() which already has the default 
name.

Can be clearer on the Current Situation, and the Problem can be stated clearly as well, with clearer references to how your above changes resulted in the test to break. For example:

`
<Current Situation> AddCommandIntegrationTest populates its model using getTypicalPersons() which contains a person with the default name. This test then creates another person using the default name, which is expected to pass. <Problem> However, the above update in equality causes this test to fail. 
`

Feel free to rephrase this better if you think mine isn't that clear :P The rest of the commit message is great!

CodyChew:
> So, I think the following segment:
> 
> > This is because AddCommandIntegrationTest uses the default name to
> > build a valid person and then adds it into an existing model. In this case,
> > the existing model uses getTypicalPersons() which already has the default
> > name.
> 
> Can be clearer on the Current Situation, and the Problem can be stated clearly as well, with clearer references to how your above changes resulted in the test to break. For example:
> 
> `<Current Situation> AddCommandIntegrationTest populates its model using getTypicalPersons() which contains a person with the default name. This test then creates another person using the default name, which is expected to pass. <Problem> However, the above update in equality causes this test to fail. `
> 
> Feel free to rephrase this better if you think mine isn't that clear :P The rest of the commit message is great!

Thanks for the tip @Zhiyuan-Amos! How does this sound?

```
Also, let's change the default name in PersonBuilder utility class. 
This is because AddCommandIntegrationTest populates its model 
using getTypicalPersons() which contains a person with the default 
name. This test then creates another person using the default name,
 to be added into the model, which is expected to pass. However,
 the above update in equality causes this test to fail. Thus we need 
to change the default name to a unique name that is not inside 
getTypicalPersons(). Also, the default email is changed for consistency 
in naming the default values, it does not affect current test cases.
```

Zhiyuan-Amos:
@CodyChew yes this is great 👍

@damithc is the commit message (above comment) clear enough?

damithc:
> @damithc is the commit message (above comment) clear enough?

Yes, I think it's good to go.
@CodyChew you can go ahead and submit the next iteration with the updated commit message.

damithc (approved):

Zhiyuan-Amos (approved):

PR #76, opened 2021-01-05
Page breaks cut through content when converted to PDF #59

### LINKED ISSUE #59: Page breaks cut through content when converted to PDF
image

MarkBind doesn't seem to have this problem. A Jekyll-specific problem?

RuiFengg:
Fixes #59 

The page breaks cut through content if no specific media query is specified

Let's fix this by applying the media print query `break-inside: avoid` on body elements 

I've tested this on my 2103 team's UG and on AB3 to verify the fix

### DIFF
--- a/docs/_sass/minima/_base.scss
+++ b/docs/_sass/minima/_base.scss
@@ -270,3 +270,12 @@ table {
             -ms-overflow-style: -ms-autohiding-scrollbar;
   }
 }
+
+/**
+ * Prevents page break to cut through content when converting to pdf
+ */
+@media print {
+  body {
+    display: block;
+  }
+}


damithc:
Thanks for the PR @RuiFengg 
Let's wait to see if anyone can review it.

wxwxwxwx9:
Hi there! I took a quick look at the problem and the proposed solution. It seems like the page-break problem stems from the use of `display: flex`. Thus, in the commit, the issue was actually solved by replacing `display: flex` (flexbox) with `display: block`. According to what I found out, it seems like flexbox doesn't go well with page-break in browsers generally. 

If you were to remove `break-inside: avoid` and keep only `display: block`, I think the problem would still be fixed. I tried it locally and it seems to be the case. Perhaps you can verify it on your side as well? 

I think a possible hotfix can be adding a print media query: 
```
@media print {
    body {
        display: block;
    }
}
```

It would be great if you can help me test out whether the hotfix would resolve all the buggy cases that you have encountered so far :-)

Regarding a proper fix -- I'm not sure why `display: flex` was used in the first place. But there might have a been a good reason for it; thus I'm not sure if it would be good to just change it to `display: block`. 

If we can ascertain that changing `display: flex` to `display: block` would not cause any regressions, then I think that would be the way to go forward.

RuiFengg:
@wxwxwxwx9 Thx for helping me check through! I tried the hotfix suggested on my end and it does solve the page break issue as well. Should we go ahead with this?

damithc:
> @wxwxwxwx9 Thx for helping me check through! I tried the hotfix suggested on my end and it does solve the page break issue as well. Should we go ahead with this?

Yes, let's try that.

RuiFengg:
> > @wxwxwxwx9 Thx for helping me check through! I tried the hotfix suggested on my end and it does solve the page break issue as well. Should we go ahead with this?
> 
> Yes, let's try that.

Alright I've updated the PR

damithc:
> Alright I've updated the PR

@RuiFengg use the [same bot you used earlier](https://canihasreview.pyokagan.com/se-edu/addressbook-level3/pull/76) to submit a new version.

damithc:
@RuiFengg there should only be one commit as this is just one change?

RuiFengg:
> @RuiFengg there should only be one commit as this is just one change?

Oops sry about that. I've updated the PR

damithc:
@se-edu/tech-team-level1I would like reviews from some of you for this PR please

Zhiyuan-Amos:
> @se-edu/tech-team-level1I would like reviews from some of you for this PR please

@damithc the `@se-edu/tech-team-level1` tag in the message that you sent above did not go through because it's missing a whitespace between `1` and `I` :P

@se-edu/tech-team-level1 tagging on prof's behalf

siangernlow:
Hi, I've tested the solution on my group's DG and it has worked fine as well. 
Some suggestions below.
1. It might be suitable to add a comment above the media print query to document its purpose. For example, a comment shown below might be useful.
```
/**
 * Prevents page breaks when converting to PDF
 * /
```
2. > Adding css media print queries will allow the pages
to be printed out without text lines being cut

Perhaps it would be better to describe the solution as `Adding a css media print query` instead since there was only a single query added in the commit.

3. > Refer to this S/0 discussion on dealing with page breaks
 https://stackoverflow.com/questions/907680/css-printing-avoiding-cut-in-half-divs-between-pages

The S/O discussion is not as relevant now since the solution has changed. Perhaps it would be more suitable to explain why it is done that way, i.e. the fact that using a media print query prevents the styles of other media types from being affected.

damithc:
thanks for the comments @siangernlow 
@RuiFengg what do you think (of the comments)?

RuiFengg:
> thanks for the comments @siangernlow
> @RuiFengg what do you think (of the comments)?

Thanks for reviewing @siangernlow
@damithc The comments were all relevant and the changes have been applied respectively. Edited the suggested code comment to be a bit more specific as well

damithc:
@wxwxwxwx9 can provide a review to this PR?

siangernlow (approved):
LGTM

wxwxwxwx9:
@damithc Hi prof, the changes look good to me! :-)

damithc:
@RuiFengg would the following commit message work? i.e., is it accurate and contain all the important information?
I shortened yours, added full stops, and used the full width (72 characters)
```
Fix page breaks cutting through content  when converted to PDF (#76)


Converting documents to PDF causes page breaks to cut through content.

As the PDF conversion is achieved by printing the Web page to a PDF,
let's specify the behaviour of the media print query to change the
display type of the 'body' component to 'block', which fixes this issue.
```

RuiFengg:
> @RuiFengg would the following commit message work? i.e., is it accurate and contain all the important information?
> I shortened yours, added full stops, and used the full width (72 characters)
> 
> ```
> Fix page breaks cutting through content  when converted to PDF (#76)
> 
> 
> Converting documents to PDF causes page breaks to cut through content.
> 
> As the PDF conversion is achieved by printing the Web page to a PDF,
> let's specify the behaviour of the media print query to change the
> display type of the 'body' component to 'block', which fixes this issue.
> ```

Yup this summarised version does contain all the important information describing and explaining the fix. Thanks for the edit! I will update the commit message

damithc:
Thanks @RuiFengg for the PR. I merged with a minor edit to the commit subject.
Thanks for your inputs @wxwxwxwx9 @siangernlow

PR #77, opened 2021-01-07
Fail gracefully if the application is not allowed to write files #73

### LINKED ISSUE #73: Fail gracefully if the application is not allowed to write files

ernestlim8:
Fixes #73 

After a command, AB3 displays an error message and exits after three
seconds.

Automatically exiting the application is more user-friendly because the
user is unable to use the application purposefully with the file write
error.

Let's check for write permissions, then display the error message to the user showing
the file that has no permissions. The application will automatically
exit gracefully after three seconds.

### DIFF
--- a/src/main/java/seedu/address/logic/LogicManager.java
+++ b/src/main/java/seedu/address/logic/LogicManager.java
@@ -21,7 +21,7 @@ import seedu.address.storage.Storage;
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: Insufficient permissions for %s";
     private final Logger logger = LogsCenter.getLogger(LogicManager.class);
 
     private final Model model;
@@ -48,7 +48,8 @@ public class LogicManager implements Logic {
         try {
             storage.saveAddressBook(model.getAddressBook());
         } catch (IOException ioe) {
-            throw new CommandException(FILE_OPS_ERROR_MESSAGE + ioe, ioe);
+            String message = ioe.getMessage();
+            throw new CommandException(String.format(FILE_OPS_ERROR_MESSAGE, message), ioe);
         }
 
         return commandResult;
--- a/src/main/java/seedu/address/ui/MainWindow.java
+++ b/src/main/java/seedu/address/ui/MainWindow.java
@@ -188,7 +188,6 @@ public class MainWindow extends UiPart<Stage> {
 
             return commandResult;
         } catch (CommandException | ParseException e) {
-            logger.info("Invalid command: " + commandText);
             resultDisplay.setFeedbackToUser(e.getMessage());
             throw e;
         }
--- a/src/test/java/seedu/address/logic/LogicManagerTest.java
+++ b/src/test/java/seedu/address/logic/LogicManagerTest.java
@@ -84,7 +84,7 @@ public class LogicManagerTest {
         Person expectedPerson = new PersonBuilder(AMY).withTags().build();
         ModelManager expectedModel = new ModelManager();
         expectedModel.addPerson(expectedPerson);
-        String expectedMessage = LogicManager.FILE_OPS_ERROR_MESSAGE + DUMMY_IO_EXCEPTION;
+        String expectedMessage = String.format(LogicManager.FILE_OPS_ERROR_MESSAGE, DUMMY_IO_EXCEPTION.getMessage());
         assertCommandFailure(addCommand, CommandException.class, expectedMessage, expectedModel);
     }
 


ernestlim8:
Fixes #73

damithc:
> Fixes #73

This should be in the first post for GitHub to close the issue when this PR is merged.

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -188,9 +192,37 @@ private CommandResult executeCommand(String commandText) throws CommandException
...
+                    logger.info("Exception occurred while waiting for thread");
+                } finally {
+                    // Execute handleExit on the JavaFx Thread
+                    Platform.runLater(() -> handleExit());
+                }
+                return null;
+            }
+        };
+
+        new Thread(accessDeniedTask).start();
```
JinHao-L: I found another alternative solution which is the [`PauseTransition`](https://docs.oracle.com/javase/8/javafx/api/javafx/animation/PauseTransition.html) class in the JavaFX library.

The following code produces the same effect and similarly does not freeze the UI. 
```
PauseTransition pause = new PauseTransition(Duration.seconds(3));
pause.setOnFinished(event -> handleExit());
pause.play();
```

I am not sure if this is more efficient but I feel that this solution might be more reader-friendly. 
What do you all think?

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -188,9 +192,37 @@ private CommandResult executeCommand(String commandText) throws CommandException
...
+                logger.info("Invalid command: " + commandText);
+            }
+
             throw e;
         }
     }
+
+    private void handleAccessDeniedException(Exception e) {
+        logger.info("Insufficient Permissions for file");
+        resultDisplay.setFeedbackToUser(e.getMessage());
```
JinHao-L: Currently, the feedback to the user seems like an unchecked exception:
> Could not save data to file: java.nio.file.AccessDeniedException: data\addressbook.json

Maybe it would be better if we rephrase it to `"Could not save data to file: Access Denied"` or `"Could not save data to file: Insufficient permission"`.

JinHao-L (commented):
Other than the comments below, the rest looks good! 👍

src/main/java/seedu/address/ui/MainWindow.java
ernestlim8: Agreed, this solution seems more user friendly. It works on my end as well.

src/main/java/seedu/address/ui/MainWindow.java
ernestlim8: I have updated the error message to be more readable for the user as per your suggestion.

`Could not save data to file: Insufficient permission for <filepath>`
The additional information of the file path allows the user to know which file path has insufficient permissions.

damithc (commented):
Thanks for the PR @ernestlim8 
I think it is enough to show a clear error message and stop operations. 
There is no need to exit automatically. Auto-exiting will make test automation harder (although we are not doing auto-testing of the GUI in AB3, it is done in AB4). Even in manual testing, it will make it harder for students to take screenshots if the error happens.

ernestlim8:
@damithc After commenting out the Pause Transition code, if there are insufficient file permissions, AB3 will display the error message for all valid commands. Is this behaviour sufficient for stopping operations?

damithc:
> @damithc After commenting out the Pause Transition code, if there are insufficient file permissions, AB3 will display the error message for all valid commands. Is this behaviour sufficient for stopping operations?

Sorry @ernestlim8 , overlooked this until now due to start-of-semester rush. Yes, I think that would work. Is the current code doing just that? I'm asking because I saw some unusual import statements.

damithc:
@se-edu/tech-team-level1 for your review

damithc:
Sorry @ernestlim8 I think this PR will progress very slowly as this is the semester time and this change is not urgent.

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -188,9 +190,19 @@ private CommandResult executeCommand(String commandText) throws CommandException
...
             return commandResult;
         } catch (CommandException | ParseException e) {
-            logger.info("Invalid command: " + commandText);
-            resultDisplay.setFeedbackToUser(e.getMessage());
+            if (e.getCause() instanceof AccessDeniedException) {
+                handleAccessDeniedException(e);
+            } else {
+                resultDisplay.setFeedbackToUser(e.getMessage());
+                logger.info("Invalid command: " + commandText);
+            }
```
damithc: See if you can get rid of this condition (checking of object type should be avoided if possible). That is, if the correct error message was put into the exception in the first place, you don't need a different behaviors here. That means we only need to change the place where the error is detected.

damithc (changes requested):

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -188,8 +189,8 @@ private CommandResult executeCommand(String commandText) throws CommandException
 
             return commandResult;
         } catch (CommandException | ParseException e) {
-            logger.info("Invalid command: " + commandText);
             resultDisplay.setFeedbackToUser(e.getMessage());
+            logger.info("Invalid command: " + commandText);
```
damithc: Any reason for this change?

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -156,6 +156,7 @@ void show() {
      */
     @FXML
     private void handleExit() {
+        logger.info("Calling exit");
```
damithc: Seems this change is not related to this PR?

damithc (commented):
Sorry for getting back so late @ernestlim8 
Some minor comments added.

damithc:
@ernestlim8 shall we finish this PR?

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -48,7 +48,8 @@ public CommandResult execute(String commandText) throws CommandException, ParseE
         try {
             storage.saveAddressBook(model.getAddressBook());
         } catch (IOException ioe) {
-            throw new CommandException(FILE_OPS_ERROR_MESSAGE + ioe, ioe);
+            String message = ioe.getMessage();
+            throw new CommandException(String.format(FILE_OPS_ERROR_MESSAGE, message), ioe);
```
damithc, line 52: Looks like the code assumes every IOException is caused by _insufficient permissions_? I don't think that is a valid assumption.

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -188,7 +188,6 @@ private CommandResult executeCommand(String commandText) throws CommandException
 
             return commandResult;
         } catch (CommandException | ParseException e) {
-            logger.info("Invalid command: " + commandText);
```
damithc, line 191: What if the command is actually invalid?

damithc (commented):
I have some concerns about the logic of this fix. Happy to be proven wrong though.

src/main/java/seedu/address/ui/MainWindow.java
ernestlim8: @damithc Seems like this log statement was already existing before this PR, its possible to split the 2 different exceptions and handle them separately, if the desired behaviour is to differentiate between invalid command and parse exception in the logs

src/main/java/seedu/address/logic/LogicManager.java
ernestlim8: @damithc Agreed, I will display the insufficient permissions when catching the AccessDeniedException instead, and use a more generalised message for other possible I/O Exceptions.

damithc:
To be continued by #172 
Thanks for this PR @ernestlim8

PR #80, opened 2021-01-09
Omit navbar of printed pdf for UG and DG

### LINKED ISSUE #35: UG/DG: Omit top nav bar from PDF version
Current: the top nav bar is included in the PDF version of the UG/DG.

Suggested: Although it is not a major problem, we can consider finding a way to omit it from the PDF version.

Notes: If we do that, we need to add the product name into the UG/DG page title. Perhaps we should do that either way?

ernestlim8:
Fixes #35 

Let's add a media query for printing, to hide the navigation bar by setting the display to none.

### DIFF
--- a/docs/_sass/minima/_base.scss
+++ b/docs/_sass/minima/_base.scss
@@ -271,6 +271,9 @@ table {
   }
 }
 
+/**
+ * Hides the navigation bar when converting to pdf
+ */
 @media print {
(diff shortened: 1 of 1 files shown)


damithc:
@ernestlim8 thanks for the PR. Can add a detailed commit message as per se-edu conventions?

PR #85, opened 2021-03-24
Update Logic component class diagram to show association to Storage

### LINKED ISSUE #84: Missing association from LogicManager to Storage in Logic class diagram
Hi, as per the code from the `LogicManager` class, there should be an association from `LogicManager` to `Storage` in the class diagram for the `Logic` component.
```
public class LogicManager implements Logic {
    private final Model model;
    private final Storage storage;
    private final AddressBookParser addressBookParser;
    ...
}
```

However, the current class diagram is missing this association:
image

w2vgd:
Fixes #84

Updated class diagram:
image


damithc:
Thanks very much for the PR @w2vgd

PR #102, opened 2021-05-27
Dependency of Storage to Model is not shown in DG #100

### LINKED ISSUE #100: DG: show Storage --> Model dependency
The Storage component depends on the Model (because `JsonAddressBookStorage.java` uses the `ReadOnlyAddressBook`). 
This dependency is not currently shown in the architecture diagram.
We can either show it in the architecture diagram or at least mention it in the text.

ernestlim8:
Fixes #100


damithc (changes requested):
Thanks for the PR @ernestlim8 
Shouldn't there be a png file too?

Commit message: a full stop is missing.

damithc:
Sorry for getting back so late @ernestlim8 
This is actually a weaker link than the other links. Perhaps a dashed arrow is more appropriate? Similar to how we show dependencies in class diagrams?

damithc (approved):
LGTM

damithc:
Thanks for the PR @ernestlim8 👍

PR #103, opened 2021-05-27
Dependency of Storage to Model is not documented #94

### LINKED ISSUE #94: Dependency from Storage to Model is not documented
https://github.com/nus-cs2103-AY2021S2/forum/issues/309

ernestlim8:
Fixes #94

### DIFF
--- a/docs/DeveloperGuide.md
+++ b/docs/DeveloperGuide.md
@@ -122,6 +122,7 @@ The `Model`,
 The `Storage` component,
 * can save `UserPref` objects in json format and read it back.
 * can save the address book data in json format and read it back.
+* depends on some classes in the `Model` component (because the `Storage` component's job is to save/retrieve objects that belong to the `Model`)
 
 ### Common classes
 


ernestlim8:
I realised that the format for the summary of the Logic component is slightly different from the other components, where it does not have this part.

```
The `Storage` component,
* can save `UserPref` objects in json format and read it back.
* can save the address book data in json format and read it back.
* depends on the Model component.
```
I think something like this might be better to match the format.
```
The `Logic` component,
* can parse commands using `AddressBookParser`
* can execute commands
* depends on some classes in the Model and Storage components (because the `Logic` component's job
is to modify objects of the `Model` class and store those objects using the `Storage` component)
```

docs/DeveloperGuide.md
```diff
@@ -122,6 +122,7 @@ The `Model`,
 The `Storage` component,
 * can save `UserPref` objects in json format and read it back.
 * can save the address book data in json format and read it back.
+* depends on the Model component.
```
damithc: The dependency itself should be visible in the architecture diagram (after #102 is merged). This PR can add further value by explaining why that dependency exists.

```suggestion
* depends on some classes in the `Model` component (because the `Storage` component's job it to save/retrieve objects belongs to the `Model`)
```

damithc (changes requested):
Thanks for the PR @ernestlim8 
Some tweaks suggested.

docs/DeveloperGuide.md
ernestlim8: @damithc I've amended this section accordingly, also I have provided a small suggestion in a comment above for the DG.

docs/DeveloperGuide.md
```diff
@@ -122,6 +122,8 @@ The `Model`,
 The `Storage` component,
 * can save `UserPref` objects in json format and read it back.
 * can save the address book data in json format and read it back.
+* depends on some classes in the `Model` component
+(because the `Storage` component's job it to save/retrieve objects belongs to the `Model`)
```
damithc:
```suggestion
(because the `Storage` component's job is to save/retrieve objects that belong to the `Model`)
```

Sorry, there was a typo in my initial suggestion that you followed.

damithc (changes requested):
Minor cosmetic issue to fix.

docs/DeveloperGuide.md
```diff
@@ -122,6 +122,8 @@ The `Model`,
 The `Storage` component,
 * can save `UserPref` objects in json format and read it back.
 * can save the address book data in json format and read it back.
+* depends on some classes in the `Model` component
+(because the `Storage` component's job is to save/retrieve objects belongs to the `Model`)
```
damithc:
```suggestion
* depends on some classes in the `Model` component (because the `Storage` component's job is to save/retrieve objects that belong to the `Model`)
```
Line wrapping not needed for .md files. 
You missed a typo suggestion from the earlier review (i.e., `belongs to` -> `that belong to`)

damithc (changes requested):
Almost there ...

damithc:
@ernestlim8 shall we finish this PR?

PR #106, opened 2021-08-23
Font Issue on M1 Mac Big Sur

### LINKED ISSUE #105: Font issue on M1 Mac Big Sur
See https://github.com/nus-cs2103-AY2122S1/forum/issues/66

ernestlim8:
Fixes #105 
The font used in AB3 (Segoe UI) is not preinstalled for M1 Mac Big Sur,
causing the text in AB3 to be illegible.

Lets change the default font to Open Sans, as it has the 3 font styles
used in AB3. Lets also embed this font to ensure all OS can access the
font.


damithc:
@se-edu/tech-team-level1 any reviews for this PR?

kingsleykuan:
Since we are embedding and redistributing the Open Sans font files, I believe that the license also needs to be included to comply with the font's open source licensing. Would it be possible to include the license file from the source the fonts were downloaded from? In this instance it would be Apache License 2.0.

damithc:
> Since we are embedding and redistributing the Open Sans font files, I believe that the license also needs to be included to comply with the open source licensing. Would it be possible to include the license file from the source the fonts were downloaded from? In this instance it would be Apache License 2.0.

Plus we have to be sure that the license allows us to redistribute under our license, which is MIT.

JoelHo:
> 
> 
> Since we are embedding and redistributing the Open Sans font files, I believe that the license also needs to be included to comply with the font's open source licensing. Would it be possible to include the license file from the source the fonts were downloaded from? In this instance it would be Apache License 2.0.

I'm not entirely sure which license its released under, their [Github Repo](https://github.com/googlefonts/opensans) says its OFL 1.1, whereas the [Google Fonts page](https://fonts.google.com/specimen/Open+Sans#license) says its Apache 2.0.

>
>
> Plus we have to be sure that the license allows us to redistribute under our license, which is MIT.

However, both licenses allow distributing it as part of another work (that is under another license) as long as the original license is distributed as well.

Perhaps the preview image for the README should be updated as well?

kingsleykuan:
> > Since we are embedding and redistributing the Open Sans font files, I believe that the license also needs to be included to comply with the open source licensing. Would it be possible to include the license file from the source the fonts were downloaded from? In this instance it would be Apache License 2.0.
> 
> Plus we have to be sure that the license allows us to redistribute under our license, which is MIT.

This is a great point, if the font was licensed under GPL I believe there would be some friction. Luckily Apache 2.0 (or OFL 1.1) and MIT are both permissive licenses and from what I can interpret there is no issue with compatibility regarding redistribution. Anyone feel free to correct me if I'm wrong though.

kouyk:
I have two observations here, firstly is that are there anyone on the team running macos big sur who can replicate the original bug verify if this works?

Another thing is I noticed that the CSS for the documentation has been modified as well, not sure if that is a good idea since open sans will likely need to be loaded via a Web font cdn.

kingsleykuan:
> Another thing is I noticed that the CSS for the documentation has been modified as well, not sure if that is a good idea since open sans will likely need to be loaded via a Web font cdn.

Good catch, modifying `docs/_sass/minima/initialize.scss` is unnecessary since it is for documentation rather than AB3 itself. I notice that there are already fonts included there that target macOS so this change is likely not needed. Cannot confirm as I do not have macOS however.

src/main/java/seedu/address/ui/UiManager.java
```diff
@@ -66,6 +66,9 @@ void showAlertDialogAndWait(Alert.AlertType type, String title, String headerTex
     private static void showAlertDialogAndWait(Stage owner, AlertType type, String title, String headerText,
                                                String contentText) {
         final Alert alert = new Alert(type);
+        alert.getDialogPane().getStylesheets().add("fonts/OpenSans-Bold.ttf");
+        alert.getDialogPane().getStylesheets().add("fonts/OpenSans-Light.ttf");
+        alert.getDialogPane().getStylesheets().add("fonts/OpenSans-SemiBold.ttf");
```
JoelHo, line 71: I'm not sure this is not how you include a font in JavaFX

JoelHo: Should this be the way to include fonts to the stylesheet?

src/main/resources/view/DarkTheme.css
```diff
@@ -218,7 +218,7 @@
     -fx-border-width: 2;
     -fx-background-radius: 0;
     -fx-background-color: #1d1d1d;
-    -fx-font-family: "Segoe UI", Helvetica, Arial, sans-serif;
+    -fx-font-family: "Open Sans", Helvetica, Arial, sans-serif;
```
JoelHo: I noticed you are using Open Sans Regular, however ```OpenSans-Regular.ttf``` seems to be missing from the PR.

docs/_sass/minima/initialize.scss
```diff
@@ -2,7 +2,7 @@
 
 // Define defaults for each variable.
 
-$base-font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Segoe UI Symbol", "Segoe UI Emoji", "Apple Color Emoji", Roboto, Helvetica, Arial, sans-serif !default;
+$base-font-family: -apple-system, BlinkMacSystemFont, "Open Sans", "Segoe UI Symbol", "Segoe UI Emoji", "Apple Color Emoji", Roboto, Helvetica, Arial, sans-serif !default;
```
JoelHo: Should the font for the documentation be changed, as per the discussion? (Thanks @kouyk @kingsleykuan !)

JoelHo (changes requested):
Building the project seems to not have the desired effect of changing the font,

image

All text seems to be displayed in the system font rather than Open Sans.

Instead, perhaps you can try including this in DarkTheme.css for all the necessary fonts
```
@font-face {
    font-family: "Open Sans";
    src: url("/fonts/OpenSans-Regular.ttf");
}
```

Perhaps the changes are working for you since you have it installed as a system font?

src/main/java/seedu/address/ui/UiManager.java
kingsleykuan: Agreeing with Joel, wouldn't `getStylesheets()` return a list for adding paths to stylesheets and not fonts? https://openjfx.io/javadoc/11/javafx.graphics/javafx/scene/Scene.html#getStylesheets()

damithc:
Once we are happy with the code, we can generate a JAR file and post it in that forum thread to get those students to verify that it works.

damithc:
@ernestlim8 can submit a new version as per comments? Also, share a JAR file created from the new code so that we can give it a try.

ernestlim8:
@JoelHo @kingsleykuan @kouyk Thanks for the inputs, I have created v2 of the PR and the jar file for your review. I disabled the Open Sans font on Font Book in my Mac, and it seems to be working properly. 
[addressbook.jar.zip](https://github.com/se-edu/addressbook-level3/files/7036439/addressbook.jar.zip)

damithc:
> @JoelHo @kingsleykuan @kouyk Thanks for the inputs, I have created v2 of the PR and the jar file for your review. I disabled the Open Sans font on Font Book in my Mac, and it seems to be working properly.
> [addressbook.jar.zip](https://github.com/se-edu/addressbook-level3/files/7036439/addressbook.jar.zip)

Thanks @ernestlim8 I tried on Windows and seems to work.

kouyk:
> @JoelHo @kingsleykuan @kouyk Thanks for the inputs, I have created v2 of the PR and the jar file for your review. I disabled the Open Sans font on Font Book in my Mac, and it seems to be working properly.
> [addressbook.jar.zip](https://github.com/se-edu/addressbook-level3/files/7036439/addressbook.jar.zip)

Confirmed to be working on Ubuntu openjdk 11, Open Sans does appear to be used although I don't have it installed.

One suggestion I have is to include all the weights of the font, such that if anyone decides to tweak the aesthetics they have more choices and won't need to source for the fonts and run into inconsistencies if they happen to download slightly different versions.

damithc:
At least one student confirmed the new JAR solves the problem.

@ernestlim8 perhaps submit another version based on @kouyk 's comment above?
Don't forget to add `Fixes #1234` annotation to the PR description too.

kouyk (commented):
I like that the screenshot has been updated, though it looks a little blurry. If I were to nitpick, it would be that the fonts while all stored within a single directory along with the license, makes it difficult to add other fonts without the license begin confused, would it be better to have a subdirectory for Open Sans?

The commit message should be imperative as well if we are abiding by the standards.

If a single commit does a single thing, should there be 2 different commits, one changing the font, and another updating the screenshot?

JoelHo:
> 
> 
> I like that the screenshot has been updated, though it looks a little blurry. 
> 


Maybe the blurry issue is to do with how MacOS renders text smoothly (ironically)? 

Really splitting hairs here but the updated screenshot does not include the ```Enter Command here...``` text, perhaps making it a bit unclear. Additionally, the window seems to have resized from defaults (or could it just be how it gets rendered on MacOS?)

Here's a screenshot from Windows with the ```Enter Command here...``` for ease of replacement

Screenshot 2021-08-24 183530

damithc:
@ernestlim8 looks like the help window still needs to be fixed. See https://github.com/nus-cs2103-AY2122S1/forum/issues/66#issuecomment-905089555

ernestlim8:
@damithc I have fixed the help window, please let me know if any other components have issues, since I am unable to reproduce the error on my end.
[addressbook3-help-window-fixed.zip](https://github.com/se-edu/addressbook-level3/files/7043071/addressbook3-help-window-fixed.zip)

damithc:
@ernestlim8 how about the suggestions given above for the screenshot and putting the font in a sub-folder?

kingsleykuan (commented):
The code changes look good, I like how the font declarations are in a separate CSS file.

The only minor nitpick I have would be that the new screenshot is much larger and wider than the original, and makes the text hard to read on the README page. Maybe you can consider using the screenshot provided by @JoelHo?

As previously mentioned by @kouyk, changes to multiple components or aspects of the codebase might better be served with multiple commits. As the current changes are still relatively small I don't think it's a big problem, however in future you might consider using multiple commits detailing the changes instead of amending a single commit and force pushing with every change.

Otherwise LGTM pending any other feedback from the students.

damithc:
> The only minor nitpick I have would be that the new screenshot is much larger and wider than the original, and makes the text hard to read on the README page. Maybe you can consider using the screenshot provided by @JoelHo?

Good point, let's fix that.


> As previously mentioned by @kouyk, changes to multiple components or aspects of the codebase might better be served with multiple commits. As the current changes are still relatively small I don't think it's a big problem, however in future you might consider using multiple commits detailing the changes instead of amending a single commit and force pushing with every change.

This is a good point too. Splitting big PRs into a series of smaller steps/commits makes it easier to review.  At the same time, based on se-edu past practice, each commit should advance the code base as a whole in a small but _complete_ step i.e., a commit should not take the code base into a incorrect/incomplete state. For instance, the tests should pass at every commit. In this case, a commit adding the fonts files without the license would take the code base into an incorrect state. By that logic, I think all changes in this PR belong together and should be in one commit, as any splitting would result in the code base going into an incorrect state at intermediate commits.

damithc:
@kingsleykuan @kouyk @JoelHo have another look at V7. Approve if you are fully OK with the code but don't hesitate to request for changes if there is even a small thing that can be improved upon. It's common for se-edu PRs to take many iterations to merge, as we aim to get rid of all nits before merging.

src/main/resources/view/HelpWindow.fxml
```diff
@@ -14,6 +15,10 @@
   </icons>
   <scene>
     <Scene>
+        <stylesheets>
```
kouyk: The indentation here seems off

src/main/resources/view/HelpWindow.fxml
```diff
@@ -14,6 +15,10 @@
...
+            <URL value="@Fonts.css"/>
+            <URL value="@HelpWindow.css"/>
```
kouyk: Not something directly related to fonts, more of a general comment for the project, why isn't the dark theme applied to the help window? If it uses the same css means it will be more consistent as well, removing for the need of an additional css that might often be overlooked given there's no inheritance whatsoever.

src/main/resources/view/DarkTheme.css
```diff
@@ -218,7 +218,7 @@
     -fx-border-width: 2;
     -fx-background-radius: 0;
     -fx-background-color: #1d1d1d;
-    -fx-font-family: "Segoe UI", Helvetica, Arial, sans-serif;
+    -fx-font-family: "Open Sans Regular", Helvetica, Arial, sans-serif;
```
kouyk: Given that our preferred font is always available, is it still necessary for the fallback fonts to be around?

Another thought I have is that could the better fix be including the fallback fonts for every element? Of course it might be a little late at this stage to change the direction of the fix, but I'm just throwing this out for everyone to chime in.

src/main/resources/view/Fonts.css
```diff
@@ -0,0 +1,14 @@
...
+
+@font-face {
+    font-family: "Open Sans SemiBold";
+    src: url("/fonts/open-sans/OpenSans-SemiBold.ttf");
+}
+
+@font-face {
+    font-family: "Open Sans Light";
+    src: url("/fonts/open-sans/OpenSans-Light.ttf");
+}
```
kouyk, line 14: Now that we also include the other font weights, should we also define them here? One concern is definitely the definition of unused entities, but on the other hand other developers trying to tweak font weights might be confused why some weights work and some doesn't especially if they are not familiar with javafx and css. An arguably better alternative is to mention it in the developer guide to provide some hints.

kouyk (changes requested):
The main hurdles in the way of merging is definitely the minor formatting issue, otherwise no major problems already. Apart from that, I do have some thoughts that the team can perhaps provide some input.

damithc:
> The main hurdles in the way of merging is definitely the minor formatting issue, otherwise no major problems already. Apart from that, I do have some thoughts that the team can perhaps provide some input.

Good points @kouyk Others, feel free to chime in. Some of those things, if we decide to go ahead with them, can be done as separate PRs in the future.

src/main/resources/view/HelpWindow.fxml
kingsleykuan: It appears that the indentation for this file was inconsistent to begin with. Some sections are indented with 2 spaces, some with 3, and this change with 4 spaces. Since it's out of scope for this PR, perhaps the change can be standardized to 2 spaces similar to `MainWindow.fxml` and the rest of the file can be fixed in a separate PR?

src/main/resources/view/DarkTheme.css
kingsleykuan: Since we are distributing the preferred font I think we can remove the fallback fonts. In any case Helvetica and Arial may be unavailable on some systems, and including a fallback to sans-serif may lead to any inconsistencies or bugs being unnoticed and unreported.

src/main/resources/view/Fonts.css
JoelHo: I think it would be better to include the other font weights, and leave a comment that they are currently unused in AB-3. Perhaps this is out of scope of the developer guide as that currently pertains to the architecture and only mentions JavaFX in passing?

Since the focus of the project is on the software engineering aspect, maybe making a new section for it may not be so appropriate as well.

src/main/resources/view/HelpWindow.fxml
kouyk: my thoughts when highlighting this is that since we are inserting something, at least the inserted portion should be formatted correctly on its own, so the open and closing tags should be at the same indentation level, currently the opening tag is misaligned with the closing tag with an offset of 1.

src/main/resources/view/DarkTheme.css
JoelHo: Seems like this stems from the initial commit where the css was borrowed from a guide, so there should not be an issue removing the fallback fonts!

ernestlim8: I agree, since the fallback fonts would not be used they can be removed. Will also fix the indentation issue as well. Thanks @kouyk for the input.

src/main/resources/view/Fonts.css
ernestlim8: I agree with @JoelHo on this issue. Shall I add in the other font weights but leave the code commented out? So it becomes clearer to the students that the rest of the weights are not being used.

kouyk: Leaving unused code commented out looks like a bad practice that shouldn't be encouraged though, while not in the scope of CS2103, [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it) is perhaps something we could take into account.

Commenting the code has the issue of not solving the problem that introduced the code in the first place, since the intention is such that anyone else can freely use other weights without having to be familiar with CSS or even know it's existence, especially when the debugger doesn't directly reveal that a missing definition in CSS is at fault, developers will still need to be aware of the commented code to utilise it.

damithc: I'm OK to add them here, for completeness sake.

src/main/resources/view/HelpWindow.fxml
damithc: Good point @kouyk . Let's tackle it separately. Create an issue for it?

damithc:
@ernestlim8 did you remove the fallback fonts in this latest version?

damithc:
@ernestlim8 reminder ...

kouyk (approved):
All outstanding issues have been resolved, LGTM!

damithc:
Thanks for the new version @ernestlim8 
Thanks for the review @kouyk 
@JoelHo @kingsleykuan can we have your decision as well?

JoelHo (approved):
LGTM. Thanks for resolving the issues!

kingsleykuan (approved):
LGTM, thanks!

damithc:
Created a new release to include this change. Thanks @ernestlim8 for the PR and everyone else who contributed @JoelHo @kingsleykuan @kouyk

damithc:
Guys, someone encountered a problem with embedded fonts. Could you follow up? https://github.com/nus-cs2103-AY2122S1/forum/issues/246

PR #109, opened 2021-09-02
Upgrade JavaFX to version 16

### LINKED ISSUE #108: Use JavaFX 16 for Duke/AB3?
@se-edu/tech-team-level1 See https://github.com/nus-cs2103-AY2122S1/forum/issues/159

I suppose OK to use later JavaFX versions as the relevant files are bundled inside the JAR file?

kouyk:
Fixes #108 by upgrading JavaFX 11 to JavaFX 16.

While this resolves any warnings caused by FXML files created by Scene Builder 16, it introduces another warning `Unsupported JavaFX configuration: classes were loaded from 'unnamed module @3136a3cc'`. This is a result of AB3 not being [modular](https://www.oracle.com/sg/corporate/features/understanding-java-9-modules.html), hence JavaFX is also not loaded as a named module. This is an [unsupported configuration](https://bugs.openjdk.java.net/browse/JDK-8256362) and the warning is introduced in JavaFX 16. This particular warning should be safe to ignore, it has been this way since the JDK 9 days.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ task coverage(type: JacocoReport) {
 
 dependencies {
     String jUnitVersion = '5.4.0'
-    String javaFxVersion = '11'
+    String javaFxVersion = '16'
 
     implementation group: 'org.openjfx', name: 'javafx-base', version: javaFxVersion, classifier: 'win'
     implementation group: 'org.openjfx', name: 'javafx-base', version: javaFxVersion, classifier: 'mac'
--- a/src/main/resources/view/CommandBox.fxml
+++ b/src/main/resources/view/CommandBox.fxml
@@ -3,7 +3,7 @@
 <?import javafx.scene.control.TextField?>
 <?import javafx.scene.layout.StackPane?>
 
-<StackPane styleClass="stack-pane" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1">
+<StackPane styleClass="stack-pane" xmlns="http://javafx.com/javafx/16" xmlns:fx="http://javafx.com/fxml/1">
   <TextField fx:id="commandTextField" onAction="#handleCommandEntered" promptText="Enter command here..."/>
 </StackPane>
 
--- a/src/main/resources/view/HelpWindow.fxml
+++ b/src/main/resources/view/HelpWindow.fxml
@@ -9,7 +9,7 @@
 <?import javafx.scene.layout.HBox?>
 <?import javafx.stage.Stage?>
 
-<fx:root resizable="false" title="Help" type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/11" xmlns:fx="http://javafx.com/fxml/1">
+<fx:root resizable="false" title="Help" type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/16" xmlns:fx="http://javafx.com/fxml/1">
   <icons>
     <Image url="@/images/help_icon.png" />
   </icons>
--- a/src/main/resources/view/MainWindow.fxml
+++ b/src/main/resources/view/MainWindow.fxml
@@ -11,7 +11,7 @@
 <?import javafx.scene.layout.StackPane?>
 <?import javafx.scene.layout.VBox?>
 
-<fx:root type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1"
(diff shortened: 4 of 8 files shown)


damithc:
Thanks for the PR @kouyk 
Yes, I guess we can ignore that warning for now.
Let us know when ready for review.

kouyk:
The changes are simple, however, one area that might be debatable is whether to update the FXML to version 16 as well. t isn't necessary to bump the version numbers, much like how some of them are still on version 8 despite us using version 11 before this.

damithc:
> The changes are simple, however, one area that might be debatable is whether to update the FXML to version 16 as well. t isn't necessary to bump the version numbers, much like how some of them are still on version 8 despite us using version 11 before this.

Would upgrading to 16 cause any regressions in the GUI?
Would keeping it as 8 conflict with SceneBuilder? e.g., Would SceneBuilder overwrite the version number?

kouyk:
> Would upgrading to 16 cause any regressions in the GUI?

No known regressions based on my own testing. Everything looks and works the same. Someone else on the team can help test it out.
[addressbook-javafx16.zip](https://github.com/se-edu/addressbook-level3/files/7103325/addressbook-javafx16.zip)

> Would keeping it as 8 conflict with SceneBuilder? e.g., Would SceneBuilder overwrite the version number?

Yes, Scene Builder will automatically switch to its own version number upon any edits.

In the process of testing this out, there is another observation that I would like to share. I edited with both the Scene Builder downloaded and integrated into Intellij (set up before the upgrade) and the external version. The version within Intellij seems to be v11 both before and after the upgrade to JavaFX 16. I do suspect that the version within Intellij is tied to the JDK version, but since there are no explicit options to control or remove the integrated Scene Builder, I cannot confirm my suspicion. If it is indeed the case, then this PR might not be necessary since the recommendation could be to edit from within Intellij.

damithc:
> No known regressions based on my own testing. Everything looks and works the same. Someone else on the team can help test it out.
> [addressbook-javafx16.zip](https://github.com/se-edu/addressbook-level3/files/7103325/addressbook-javafx16.zip)

@se-edu/tech-team-level1 can you guys check this?

> In the process of testing this out, there is another observation that I would like to share. I edited with both the Scene Builder downloaded and integrated into Intellij (set up before the upgrade) and the external version. The version within Intellij seems to be v11 both before and after the upgrade to JavaFX 16. I do suspect that the version within Intellij is tied to the JDK version, but since there are no explicit options to control or remove the integrated Scene Builder, I cannot confirm my suspicion. If it is indeed the case, then this PR might not be necessary since the recommendation could be to edit from within Intellij.

Thanks for this update @kouyk Yes, we can't be sure what's the reason the plugin is at 11. Might change without warning. In this case we'll have to update the JavaFX tutorial as that is where we ask students to download the SceneBuilder.

JinHao-L:
> No known regressions based on my own testing. Everything looks and works the same. Someone else on the team can help test it out.
[addressbook-javafx16.zip](https://github.com/se-edu/addressbook-level3/files/7103325/addressbook-javafx16.zip)

Aside from the previously mentioned warning message about `unnamed module`, everything looks and works the same. Tested it on Ubuntu 20.04, OpenJDK 11.

The changes in the PR looks good to me. Just a slight nitpick is that I think that we should add the rationale for the change in the commit message, following the [commit standard](https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message)

JoelHo:
> 
> 
> No known regressions based on my own testing. Everything looks and works the same. Someone else on the team can help test it out.
> [addressbook-javafx16.zip](https://github.com/se-edu/addressbook-level3/files/7103325/addressbook-javafx16.zip)

Works and looks fine on my machine (Windows 11, Oracle JDK 11.0.11), same warning. 

Agree with @JinHao-L about the commit message, minor nitpick but would be good to include as part of the version bump.

damithc:
Given some teams have set up the team repo already, let's merge this later, in time for the next semester. For this semester, we can advise students to do the upgrade themselves and refer them to this PR.

kouyk:
Given that we have narrowed down the cause down to the m1 macs, I'm in favour of closing this PR since it is a workaround that doesn't resolve the root cause yet introduces more warnings.

damithc:
> Given that we have narrowed down the cause down to the m1 macs, I'm in favour of closing this PR since it is a workaround that doesn't resolve the root cause yet introduces more warnings.

Yup, let's close it. We can reopen it later if needed.

PR #115, opened 2021-12-03
Update all class diagrams which uses interfaces to non-italics

### LINKED ISSUE #89: Class diagrams: use non-italics for interfaces
See https://github.com/nus-cs2103-AY2021S2/forum/issues/276

seanjyjy:
Fixes #89 

All updated class diagrams:

ComponentManagers
LogicClassDiagram
ModelClassDiagram
ParserClasses
StorageClassDiagram
UiClassDiagram


damithc:
Thanks for the PR  @seanjyjy 
I'm wondering if this change is worth the extra verbosity in the code or if there is a less-verbose way to get rid of the italics. 
@se-edu/tech-team-level1 your thoughts?

seanjyjy:
@damithc I could use alias if thats better. 

Code example

Original
```
package Logic {
Class "<<Interface>>\nLogic"
Class LogicManager
}
...
HiddenOutside ..> "<<Interface>>\nLogic"

LogicManager .up.|> "<<Interface>>\nLogic"
```


```
package Logic {
Class "<<Interface>>\nLogic" as Logic
Class LogicManager
}

...
HiddenOutside ..> Logic

LogicManager .up.|> Logic
```

So the only code change would be changing the original line of interface to be a class with the alias

docs/diagrams/ModelClassDiagram.puml
```diff
@@ -5,16 +5,12 @@ skinparam arrowColor MODEL_COLOR
...
-Interface Model <<Interface>>
+Class "<<Interface>>\nReadOnlyAddressBook" as ReadOnlyAddressBook
+Class "<<Interface>>\nReadOnlyUserPrefs" as ReadOnlyUserPrefs
+Class "<<Interface>>\nModel" as Model
 Class AddressBook
-Class ReadOnlyAddressBook
-Class Model
 Class ModelManager
 Class UserPrefs
-Class ReadOnlyUserPrefs
```
seanjyjy, line 16: Removing these 3 field as they are unused

LimJunxue:
I think Sean's change from the Interface to Class keyword is necessary.

You can even do:
### uml implementation files
`Class Logic <<Interface>>` 
### style.puml
```
skinparam Class {
    ...
    StereotypeFontStyle plain
}
``` 
This does the same thing but possibly more maintainable.

seanjyjy:
@LimJunxue thanks for the great suggestion. Though there is a slight difference in terms of the image, but the interface word is indeed non-italics with the changes. 

Changes made via `style.puml`:

Screenshot 2021-12-09 at 2 54 20 PM

Changes made via `using alias`:

ComponentManagers

Do let me know which is the prefer method

LimJunxue:
@seanjyjy oh yea my previous comment meant that you still have to change the Interface keyword to Class, in the uml implementation files such as ComponentManagers.puml, for the class name to appear non-italicized, since apparently Interface skinparams don't work in class diagrams (https://plantuml-documentation.readthedocs.io/en/latest/formatting/all-skin-params.html#interface).

kingsleykuan:
While I think both approaches are good, I prefer the alias method proposed by @seanjyjy as it is more consistent with how abstract classes are currently done, for example:

`Class "{abstract}\nCommand" as Command`
`Class "<<Interface>>\nLogic" as Logic`

Since `{abstract}` cannot be done using PlantUML stereotypes, I believe standardising both interfaces and abstract classes to use alias would be the preferred option.

seanjyjy:
I agree with @kingsleykuan . My thought process was to use similar syntax that are available. Changing the "style" in `style.puml` might not be desirable as I am not sure how many students will bother looking at the `style.puml` at all and might assume that's the default!

@damithc Do you have any suggestion to chip in on this situation?

damithc:
> @damithc Do you have any suggestion to chip in on this situation?

@seanjyjy The current version doesn't seem _that_ verbose. The initial version was more verbose, right? Or I remembered it wrongly?

seanjyjy:
@damithc Yes you're right. The initial version was way more verbose as I had to manually changed each of the interfaces into the long text of `class ...` With the help of alias, we just have to change the main reference to add the alias and it will work fine.

damithc:
> @damithc Yes you're right. The initial version was way more verbose as I had to manually changed each of the interfaces into the long text of `class ...` With the help of alias, we just have to change the main reference to add the alias and it will work fine.

Good. No objections to the current version from me. Let's wait for a few PR reviews from @se-edu/tech-team-level1 members first.

JinHao-L (commented):
I agree with the changes made here. I don't think there is any way to remove the italic font style while keeping the `Interface` phrase intact.

I did some research on my own and it seems that the italic font styling is hardcoded and cannot be modified.
Relevant link to the issue: https://forum.plantuml.net/14983/possible-normal-interface-names-rather-italics-class-diagrams

LimJunxue (approved):
I'm good with the current changes as well :)

damithc:
@se-edu/tech-team-level1 any further views on this PR? If You are OK, provide an 'approve' review.

JinHao-L (approved):
LGTM 👍

JoelHo (commented):
Perhaps a tiny nitpick, `Interface` is capitalized whereas `abstract` is not. Should all of them be lowercase to match the standard? LGTM otherwise!

damithc:
> Perhaps a tiny nitpick, `Interface` is capitalized whereas `abstract` is not. Should all of them be lowercase to match the standard? LGTM otherwise!

Good point @JoelHo 
@seanjyjy Can you change it to lower case and push a new version?

seanjyjy:
Made the relevant changes and rebased with master

kouyk:
From the commit history it looks like something went wrong with the rebase.

damithc:
@seanjyjy can you fix the merge conflict?

seanjyjy:
@kouyk and @damithc . Yup sorry my bad, didnt notice the conflict

damithc:
Thanks for the fix @seanjyjy 
Thanks to reviewers too!

PR #116, opened 2021-12-03
Update Ui of help window to dark theme

### LINKED ISSUE #107: Inconsistent themes between main window and help window
The main window of the application is in dark mode, but the help window has a light theme to it. This is inconsistent and affects both the user experience and also result in more deviations if one were to tweak the appearance of the main window further.

The theme of the [main window](https://github.com/se-edu/addressbook-level3/blob/master/src/main/resources/view/MainWindow.fxml) is defined in [`src/main/resources/view/DarkTheme.css`](https://github.com/se-edu/addressbook-level3/blob/master/src/main/resources/view/DarkTheme.css) . However, the help window does not utilize the same stylesheets, and thus uses the default color scheme in JavaFX.

A proposed fix would be to import the same stylesheets used by the main window into the help window, since elements should all look similar regardless of the window type. An alternative would be to have a dedicated stylesheet for the help window that inherits most of the dark theme styles to allow for independent customisation.

seanjyjy:
Fixes #107 

> A proposed fix would be to import the same stylesheets used by the main window into the help window, since elements should all look similar regardless of the window type. 

Since HelpWindow has its own dedicated CSS file, it will make more sense to edit it instead of importing the same stylesheets used by main window also it doesn't work simply to import the main window's stylesheet.

> An alternative would be to have a dedicated stylesheet for the help window that inherits most of the dark theme styles to allow for independent customisation.

Personally, I feel that there isn't much UI involved in AB3 itself, so it feels a bit unnecessary to do such an abstraction. In fact, it can be encouraged for the students who plan to so as an extra improvement when taking over it as a brownfield project.

Small Changes to HelpWindow's UI to fit the DarkTheme
Screenshot 2021-12-04 at 1 42 57 AM

### DIFF
--- a/src/main/resources/view/HelpWindow.css
+++ b/src/main/resources/view/HelpWindow.css
@@ -1,3 +1,20 @@
 #copyButton, #helpMessage {
     -fx-font-family: "Open Sans";
+    -fx-text-fill: white;
+}
+
+#copyButton {
+    -fx-background-color: dimgray;
+}
+
+#copyButton:hover {
+    -fx-background-color: gray;
+}
+
+#copyButton:armed {
+    -fx-background-color: darkgray;
+}
+
+#helpMessageContainer {
+    -fx-background-color: derive(#1d1d1d, 20%);
 }
--- a/src/main/resources/view/HelpWindow.fxml
+++ b/src/main/resources/view/HelpWindow.fxml
@@ -20,7 +20,7 @@
         <URL value="@HelpWindow.css" />
       </stylesheets>
 
-      <HBox alignment="CENTER">
+      <HBox alignment="CENTER" fx:id="helpMessageContainer">
         <children>
           <Label fx:id="helpMessage" text="Label">
             <HBox.margin>


damithc:
Thanks for the PR @seanjyjy How about the `Copy URL` button? 
@se-edu/tech-team-level1 for your inputs please.

seanjyjy:
@damithc Not very sure what design would best fit it. However, currently, it doesn't feel out of place in my opinion. However, if @se-edu/tech-team-level1 has any suggestions feel free to give some inputs

kouyk:
I tried out a simple color scheme for the copy button, see screenshots below for how it looks. The button stands out less and fits the overall theme better.

Normal:
default

Hover:
hover

Pressed:
pressed

[These](https://github.com/se-edu/addressbook-level3/files/7691650/copybutton.patch.txt) are the changes I've made if you want to try it out.

Edit: switched to the `armed` instead of `pressed` as the latter only works for mouse input.

kingsleykuan:
Tested the changes proposed by @kouyk and they look good! I prefer swapping the colours for `hover` and `armed`, however that is just personal preference. Perhaps they could be swapped? Only tested on Ubuntu however it should be the same on other platforms.

seanjyjy:
Thanks @kouyk and @kingsleykuan for the suggestion. The color scheme provided by @kouyk is great. I followed the suggestion from @kingsleykuan as it seems more intuitive. However, I do not have much experience in ui/ux so i am open for more suggestion if needed

kouyk (approved):
I introduced the various colors to make up for default button effects that become disabled once we try to style the button. My personal intuition are that as buttons are pressed down they become in the shadow and hence darker, but that's likely to be personal preference. The application does look consistent either way.

damithc:
Will merge after receiving one more approving review. @kingsleykuan are you OK with the current code?

kingsleykuan (approved):
I'm okay with the style and code, looks good to merge :thumbsup:

damithc:
@seanjyjy Thanks again for the fix.
Thanks to the reviewers @kouyk @kingsleykuan too.

PR #119, opened 2021-12-28
Revert "Font Issue on M1 Mac Big Sur"

### LINKED ISSUE #111: Loading of embedded fonts via CSS causes runtime issues
When loading fonts via CSS, it is required that the absolute path to the font does not contain any spaces. Spaces will be encoded as `%20` in the path and passed to the loading function directly, resulting in a failure.

Reference: https://stackoverflow.com/questions/33973921/javafx-font-face-css-error-loadstylesheetunprivileged/41753098

This bug occurs when the project is launched directly in the IDE, and when the project directory or any of the parent directories contains a space in its name. It does not affect any JAR files.

kouyk:
Reverts se-edu/addressbook-level3#106 to resolve #111.

The original PR aimed to side step the garbled fonts on M1 Mac devices by embedding another font of our choice. One upside and unintended consequence is that fonts are consistent across the various OS regardless of font availability, helping to avoid any potential UI bugs such as text being cut off prematurely etc. However, this does bring about unnecessary complication as more resources are being loaded into the application, such as the one surfaced [here](https://github.com/nus-cs2103-AY2122S1/forum/issues/246), mandating the use of even more workarounds such as the one referenced in #111.

The M1 MacOS issue was eventually resolved through the use of an alternative OpenJDK build that happened to have native M1 binaries as well. Given that the issue originated from OpenJDK rather than AB3 itself, it no longer make sense for the "fix" to be applied to AB3, which seems to cause more trouble than it is worth. Given that most students are fresh to JavaFX, the additional complication does not seem warranted.


damithc:
@kouyk This is an automatic git revert, correct? And this requires us to release a new version, right?

kouyk:
> @kouyk This is an automatic git revert, correct? And this requires us to release a new version, right?

Yes, that's right. I reverted through github, I might need to do some manual fixes to resolve the conflicts now that the help window PR has been merged in.

damithc:
Thanks very much for the fix @kouyk 
I'll do a new release

PR #129, opened 2022-03-28
Test PR

pyokagan:
My hypothesis is that GitHub is treating any ref named "head" (case-insensitive) specially. Let's see if that is the case.


pyokagan:
Nope no dice :-( Seems like I really do need to add the workflow permission.

Issue #122, opened 2022-02-03
Future-proof code for migration to Java 17


yhtMinceraft1010X:
In light of a potential migration to Java 17 in the future, one suggestion I would like to make is to address these warnings from IntelliJ:
![Screenshot (69)](https://user-images.githubusercontent.com/69678785/152359639-42e13044-5edf-49ff-9f61-cf68a4961340.png)

baskargopinath:
i dont see these errors on my intellij when configuring the project for J17, have they been fixed? if not how do u make them appear?

ok found out on mac intellij
Code -> Inspect Code -> Whole Project -> Analyze -> Click on **Java language level migration aids**

PR #138, opened 2022-06-23
Upgrade to Java 17

### LINKED ISSUE #137: Upgrade to Java 17

fans2619:
Fixes #137

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -29,10 +29,10 @@ jobs:
       - name: Validate Gradle Wrapper
         uses: gradle/wrapper-validation-action@v1
 
-      - name: Setup JDK 11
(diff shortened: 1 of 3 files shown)


fans2619:
@damithc The app is runnable in Java 17 when simply changing the Java version in `build.gradle` and actually I don't see big problems.🤔 There are just some minor issues as pointed out by IDE and also in #122 . I'll look into those later.

image

damithc:
@fans2619 That's good news. We also want to make sure relevant guides at https://se-education.org/guides/ are compatible with Java 17. e.g., JUnit tutorial  JavaFX tutorial

We also need to decide if we should transition the project to use Java modules

@se-edu/tech-team-level1 What do you guys think?

damithc:
Putting this hold for now.

Issue #126, opened 2022-03-20
Changing the log level in config.json does not work


damithc:
See https://github.com/nus-tic4002-AY2122s2/forum/issues/46

damithc:
@se-edu/tech-team-level1 can confirm if this is a bug?

vimuthm:
I seem to face the same issue, but only with `FINE` and further depth log levels. (`WARNING` seems to work)

Issue seems to be stemming from the fact that in `MainApp.java`, a logger is initialized as a class attribute first before the `config.json` file is read within the `init()` method of the same file, therefore defaulting the log level to `INFO` and there is no subsequent overwriting because of the null check here in `LogsCenter.java`;

```
private static void addConsoleHandler(Logger logger) {
        if (consoleHandler == null) {
            consoleHandler = createConsoleHandler();
        }
        logger.addHandler(consoleHandler);
}
```

So only log levels higher than `INFO` seem to work. 
I'm unsure of the best practices concerned here with how and when to set the log level.

edit: Manually changing the default log level inside `LogsCenter.java` doesn't fix the issue. So something else might be causing the issue

LimJunxue:
For me, changing the log level in `config.json` to all 4 logging levels as stated in the logging conventions does not limit the logged messages at all and remains at `INFO`. 

I agree with @VimuthM on the source of the issue. Calling `setLevel` of the `consoleHandler` and `fileHandler` in `addConsoleHandler` and `addFileHandler` after the null check like so will result in the correct outcome shown below when setting the log level to `WARNING` in `config.json`.
```
private static void addConsoleHandler(Logger logger) {
        if (consoleHandler == null) {
            consoleHandler = createConsoleHandler();
        }
        consoleHandler.setLevel(currentLogLevel);
        logger.addHandler(consoleHandler);
}
```
![image](https://user-images.githubusercontent.com/60027105/159171315-9e7616a6-bc49-490c-965e-860a41a5e5c8.png)
All subsequent log messages after that last line lower than `WARNING` will be discarded, which is the expected outcome. Also, not sure if the intent is to show the `currentLogLevel` with an `INFO` message before applying the log level of the `config.json`.

Py0000:
I seem to be facing the same issues as mentioned and similar to @VimuthM, only `FINE` seems to have the issue. 

`WARNING` and `SEVERE` seems to be working as intended after changing the logLevel in `config.json` as well as the `currentLogLevel` field in `LogsCenter.java`.   

[...]

PR #152, opened 2023-01-03
Rephrase 'file not found' logs shown during first run #139

### LINKED ISSUE #139: Rephrase 'file not found' logs shown during first run

image

LimJunxue:
Fixes #139 
Result:

image

### DIFF
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -117,6 +117,9 @@ public class MainApp extends Application {
 
         try {
             Optional<Config> configOptional = ConfigUtil.readConfig(configFilePathUsed);
+            if (!configOptional.isPresent()) {
+                logger.info("Config file not found. Will be starting with a new default config file.");
+            }
             initializedConfig = configOptional.orElse(new Config());
         } catch (DataConversionException e) {
             logger.warning("Config file at " + configFilePathUsed + " is not in the correct format. "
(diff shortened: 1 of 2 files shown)


damithc:
@LimJunxue the log messages that caused the initial confusion are still there. Can we _replace_  (rather than add onto) those messages with the new messages?

LimJunxue:
@damithc hmm the `Json file ... not found.` log comes from `JsonUtil#readJsonFile` which does not create the new files though. Will it be better if we change the `Config file not found. Will be starting with a new default config file.` lines to `Config file not yet **created**. Will be starting with a new default config file.` while still keeping the JsonUtil log?

damithc:
> @damithc hmm the `Json file ... not found.` log comes from `JsonUtil#readJsonFile` which does not create the new files though.

I see. We can try lowering the log level of the initial log message e.g., `logger.fine` so that it appears only for developers specifically running the app at a lower lower log level (assuming the default log level is `info`). If that doesn't work, we can also consider removing it altogether. What do you think?

LimJunxue:
The changing of log level issue is still unfixed which might cause some confusion for developers. So I think removing the log is fine too because it is not needed when the added logs covers all uses of `JsonUtil#readJsonFile` in this project.

damithc:
> The changing of log level issue is still unfixed which might cause some confusion for developers.

Are you sure? When I changed the line as follows, it doesn't appear in the console anymore.

```java
logger.fine("Json file " + filePath + " not found");
```

But you are right in that developers might be confused if they don't know that `fine` level log messages don't appear by default.

Another option is to report when the file is found, rather than not found:
```java
            return Optional.empty();
        }
        logger.info("Json file " + filePath + " found");
```

LimJunxue:
> The changing of log level issue is still unfixed which might cause some confusion for developers.

This was referring to #126 

The result will be this, is this ok?
image

damithc:
> The result will be this, is this ok?
> image

@LimJunxue This one is OK. The more interesting one is the case of the file not being found. That's the one that confused students.

damithc:

Hikoya (approved):
LGTM. The new logs are much more clearer to understand.

damithc:
Sorry, it took a while for me to come back to this PR.
I'm wondering if including the target filename in the log message is better (rather than the current `Prefs file not found. ...`, `Preferences file prefs.json not found. ...`)

damithc:
Closing in favor of #170

PR #164, opened 2023-05-11
[#139] Rephrase 'file not found' logs shown during first run

### LINKED ISSUE #139: Rephrase 'file not found' logs shown during first run

Eclipse-Dominator:
Fixes #139 
Continued from #152 

Modified warning messages to show filepath

### DIFF
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -117,6 +117,9 @@ public class MainApp extends Application {
 
         try {
             Optional<Config> configOptional = ConfigUtil.readConfig(configFilePathUsed);
+            if (!configOptional.isPresent()) {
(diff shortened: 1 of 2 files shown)


Eclipse-Dominator:
While working on this, I realised that the 
```java
        } catch (IOException e) {
            logger.warning("Problem while reading from the file " + storage.getAddressBookFilePath()
                    + ". Will be starting with an empty AddressBook");
        }
```

and 

```java
        } catch (IOException e) {
            logger.warning("Problem while reading from the file " + prefsFilePath
                    + ". Will be starting with an empty AddressBook");
        }
```
is unnecessary as the actual implementation of the methods used in the try block already handles the IO Exception, I will link a separate issue here #167

Eclipse-Dominator:
use #170

PR #165, opened 2023-05-11
[#73] no write access fail gracefully

### LINKED ISSUE #73: Fail gracefully if the application is not allowed to write files

Eclipse-Dominator:
Fixes #73 
Continued from #77 

Renames warning message to remind user to exit the application immediately.

### DIFF
--- a/src/main/java/seedu/address/logic/LogicManager.java
+++ b/src/main/java/seedu/address/logic/LogicManager.java
@@ -21,7 +21,7 @@ import seedu.address.storage.Storage;
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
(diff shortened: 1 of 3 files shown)


Eclipse-Dominator:
There is an issue with the exit command where exit will attempt to save all file before exiting and due to the permission error, it will prevent the program from exiting. 
 
I came up with the following solutions regarding the above
1. Immediately exit regardless of save errors
2. Adds a 3s timer to show the fail to exit before exiting (only on exit command)
3. Change warning message to tell the user to close via clicking on the x button [current approach]

I am not sure which solution is the right approach.

---- 
A separate issue is when the preference folder is not writeable. Error messages like so can still be seen
image
The IO exception is still caught here
```java
        //Update config file in case it was missing to begin with or there are new/unused fields
        try {
            ConfigUtil.saveConfig(initializedConfig, configFilePathUsed);
        } catch (IOException e) {
            logger.warning("Failed to save config file : " + StringUtil.getDetails(e));
        }
        return initializedConfig;
```
So I am not sure if we should change this message.

Eclipse-Dominator:
Refer to #172

Issue #167, opened 2023-05-11
Unnecessary catch block in MainApp and Exceptions in Interfaces


Eclipse-Dominator:
In MainApp.java,

```java
        try {
            addressBookOptional = storage.readAddressBook();
            if (!addressBookOptional.isPresent()) {
                logger.info("Data file not found. Will be starting with a sample AddressBook");
            }
            initialData = addressBookOptional.orElseGet(SampleDataUtil::getSampleAddressBook);
        } catch (DataConversionException e) {
            logger.warning("Data file not in the correct format. Will be starting with an empty AddressBook");
            initialData = new AddressBook();
        } catch (IOException e) {
            logger.warning("Problem while reading from the file. Will be starting with an empty AddressBook");
            initialData = new AddressBook();
        }
```
```java
        try {
            Optional<UserPrefs> prefsOptional = storage.readUserPrefs();
            initializedPrefs = prefsOptional.orElse(new UserPrefs());
        } catch (DataConversionException e) {
            logger.warning("UserPrefs file at " + prefsFilePath + " is not in the correct format. "
                    + "Using default user prefs");
            initializedPrefs = new UserPrefs();
        } catch (IOException e) {
            logger.warning("Problem while reading from the file. Will be starting with an empty AddressBook");
            initializedPrefs = new UserPrefs();
        }
```

The IOException was due to `readUserPrefs()` and `readAddressBook()` possibly throwing `IOException` suggested by their respective interface files.

However, in the actual implementation of the methods, the `IOException` is actually already handled within the functions itself so this means that `IOException` is actually never thrown. Thus, I believe it is safe to remove the catch statement for `IOException` as well as in the interface. 

e.g. 
### readUserPref()
In StorageManager.java
```java
    @Override
    public Optional<UserPrefs> readUserPrefs() throws DataConversionException, IOException {
        return userPrefsStorage.readUserPrefs();
    }
```
In JsonUserPrefStorage.java
```java

    @Override
    public Optional<UserPrefs> readUserPrefs() throws DataConversionException {
        return readUserPrefs(filePath);
    }

    /**
     * Similar to {@link #readUserPrefs()}
     * @param prefsFilePath location of the data. Cannot be null.
     * @throws DataConversionException if the file format is not as expected.
     */
    public Optional<UserPrefs> readUserPrefs(Path prefsFilePath) throws DataConversionException {
        return JsonUtil.readJsonFile(prefsFilePath, UserPrefs.class);
    }
```
[...]

PR #170, opened 2023-05-11
[#139] Rephrase 'file not found' during first run

### LINKED ISSUE #139: Rephrase 'file not found' logs shown during first run

Eclipse-Dominator:
Fixes #139 
Continued from #152 

Modified warning messages to show filepath
image
image

### DIFF
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -79,14 +79,17 @@ public class MainApp extends Application {
         try {
             addressBookOptional = storage.readAddressBook();
             if (!addressBookOptional.isPresent()) {
-                logger.info("Creating a new data file populated with a sample AddressBook");
+                logger.info("Creating a new data file " + storage.getAddressBookFilePath()
+                        + " populated with a sample AddressBook.");
             }
             initialData = addressBookOptional.orElseGet(SampleDataUtil::getSampleAddressBook);
         } catch (DataConversionException e) {
-            logger.warning("Data file not in the correct format. Will be starting with an empty AddressBook");
+            logger.warning("Data file at " + storage.getAddressBookFilePath() + " is not in the correct format."
+                    + " Will be starting with an empty AddressBook.");
             initialData = new AddressBook();
         } catch (IOException e) {
-            logger.warning("Problem while reading from the file. Will be starting with an empty AddressBook");
+            logger.warning("Problem while reading from the file " + storage.getAddressBookFilePath()
+                    + ". Will be starting with an empty AddressBook.");
             initialData = new AddressBook();
         }
 
@@ -118,12 +121,12 @@ public class MainApp extends Application {
         try {
             Optional<Config> configOptional = ConfigUtil.readConfig(configFilePathUsed);
             if (!configOptional.isPresent()) {
-                logger.info("Creating new config file.");
+                logger.info("Creating new config file " + configFilePathUsed);
             }
             initializedConfig = configOptional.orElse(new Config());
         } catch (DataConversionException e) {
-            logger.warning("Config file at " + configFilePathUsed + " is not in the correct format. "
-                    + "Using default config properties");
+            logger.warning("Config file at " + configFilePathUsed + " is not in the correct format."
+                    + " Using default config properties.");
             initializedConfig = new Config();
         }
 
@@ -143,21 +146,22 @@ public class MainApp extends Application {
      */
     protected UserPrefs initPrefs(UserPrefsStorage storage) {
         Path prefsFilePath = storage.getUserPrefsFilePath();
-        logger.info("Using prefs file : " + prefsFilePath);
+        logger.info("Using preference file : " + prefsFilePath);
 
         UserPrefs initializedPrefs;
         try {
             Optional<UserPrefs> prefsOptional = storage.readUserPrefs();
             if (!prefsOptional.isPresent()) {
-                logger.info("Creating new prefs file.");
+                logger.info("Creating new preference file " + prefsFilePath);
             }
             initializedPrefs = prefsOptional.orElse(new UserPrefs());
         } catch (DataConversionException e) {
-            logger.warning("UserPrefs file at " + prefsFilePath + " is not in the correct format. "
-                    + "Using default user prefs");
+            logger.warning("Preference file at " + prefsFilePath + " is not in the correct format."
+                    + " Using default preferences.");
             initializedPrefs = new UserPrefs();
         } catch (IOException e) {
-            logger.warning("Problem while reading from the file. Will be starting with an empty AddressBook");
+            logger.warning("Problem while reading from preference file " + prefsFilePath
+                    + ". Will be starting with an empty AddressBook.");
             initializedPrefs = new UserPrefs();
         }
 
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -79,7 +79,7 @@ public class MainApp extends Application {
         try {
             addressBookOptional = storage.readAddressBook();
             if (!addressBookOptional.isPresent()) {
-                logger.info("Data file not found. Will be starting with a sample AddressBook");
+                logger.info("Creating a new data file populated with a sample AddressBook");
             }
             initialData = addressBookOptional.orElseGet(SampleDataUtil::getSampleAddressBook);
         } catch (DataConversionException e) {
@@ -117,6 +117,9 @@ public class MainApp extends Application {
 
         try {
             Optional<Config> configOptional = ConfigUtil.readConfig(configFilePathUsed);
+            if (!configOptional.isPresent()) {
+                logger.info("Creating new config file.");
+            }
             initializedConfig = configOptional.orElse(new Config());
         } catch (DataConversionException e) {
             logger.warning("Config file at " + configFilePathUsed + " is not in the correct format. "
@@ -145,6 +148,9 @@ public class MainApp extends Application {
         UserPrefs initializedPrefs;
         try {
             Optional<UserPrefs> prefsOptional = storage.readUserPrefs();
+            if (!prefsOptional.isPresent()) {
+                logger.info("Creating new prefs file.");
+            }
             initializedPrefs = prefsOptional.orElse(new UserPrefs());
         } catch (DataConversionException e) {
             logger.warning("UserPrefs file at " + prefsFilePath + " is not in the correct format. "
--- a/src/main/java/seedu/address/commons/util/JsonUtil.java
+++ b/src/main/java/seedu/address/commons/util/JsonUtil.java
@@ -60,9 +60,9 @@ public class JsonUtil {
         requireNonNull(filePath);
 
         if (!Files.exists(filePath)) {
-            logger.info("Json file " + filePath + " not found");
             return Optional.empty();
         }
+        logger.info("JSON file " + filePath + " found.");
 
         T jsonFile;
 


Eclipse-Dominator:
While working on this, I realised that the 
```java
        } catch (IOException e) {
            logger.warning("Problem while reading from the file " + storage.getAddressBookFilePath()
                    + ". Will be starting with an empty AddressBook");
        }
```

and 

```java
        } catch (IOException e) {
            logger.warning("Problem while reading from the file " + prefsFilePath
                    + ". Will be starting with an empty AddressBook");
        }
```
is unnecessary as the actual implementation of the methods used in the try block already handles the IO Exception, I have made a separate issue regarding this: #167

src/main/java/seedu/address/commons/util/JsonUtil.java
```diff
@@ -60,9 +60,9 @@ public static <T> Optional<T> readJsonFile(
         requireNonNull(filePath);
 
         if (!Files.exists(filePath)) {
-            logger.info("Json file " + filePath + " not found");
             return Optional.empty();
         }
+        logger.info("Json file " + filePath + " found");
```
damithc: Json -> JSON?

src/main/java/seedu/address/MainApp.java
```diff
@@ -140,18 +147,24 @@ protected Config initConfig(Path configFilePath) {
...
         Path prefsFilePath = storage.getUserPrefsFilePath();
-        logger.info("Using prefs file : " + prefsFilePath);
+        logger.info("Using preference file : " + prefsFilePath);
 
         UserPrefs initializedPrefs;
         try {
             Optional<UserPrefs> prefsOptional = storage.readUserPrefs();
+            if (!prefsOptional.isPresent()) {
+                logger.info("Preference file " + prefsFilePath
+                        + " not found. Will be starting with a new default preference file.");
```
damithc: The 'not found' part still feels like an error. I'm wondering if we should just do something like this instead?
```suggestion
                logger.info("Creating a new new preference file " + prefsFilePath);
```

damithc (changes requested):
I added a few points to consider.
Perhaps the first two commits can be combined, as the two changes make more sense when viewed together?

src/main/java/seedu/address/MainApp.java
Eclipse-Dominator: I think it could be the case where the file is missing/deleted where the other files exist. I think it really depends on if whether the previous files already exist. 

e.g. 
if config exist, but not preference and data before launch
or if config, preference exist but not data before launch

Should we just standardize all of this to "creating a new XXX file..." or make it a separate case should certain files already exist prior to this.

Another problem is when the file does not have read access, the IOException is caught in the read function so the program will think that the file does not exist (even though it is due to a lack of permission to read). I think this should be solved in a separate PR since it is a separate issue.

damithc: Good point @Eclipse-Dominator. 
There is also the possibility that all three files are deleted, in which case we can't differentiate between new vs deleted. So, I guess we can simply standardize to "creating a new XXX file..." 

>Another problem is when the file does not have read access, the IOException is caught in the read function so the program will think that the file does not exist (even though it is due to a lack of permission to read). I think this should be solved in a separate PR since it is a separate issue.

Yes, we can address that separately.

Eclipse-Dominator: I have updated the logs message and you can refer to them in the screenshot in the PR.

damithc: Looks good @Eclipse-Dominator 
You can go ahead and submit the next iteration.

src/main/java/seedu/address/commons/util/JsonUtil.java
```diff
@@ -60,9 +60,9 @@ public static <T> Optional<T> readJsonFile(
         requireNonNull(filePath);
 
         if (!Files.exists(filePath)) {
-            logger.info("Json file " + filePath + " not found");
             return Optional.empty();
         }
+        logger.info("JSON file " + filePath + " found");
```
damithc: Add full stop, to be consistent with other log messages? Applies to other places as well.

src/main/java/seedu/address/MainApp.java
```diff
@@ -143,21 +146,23 @@ protected Config initConfig(Path configFilePath) {
...
-                logger.info("Prefs file not found. Will be starting with a new default prefs file.");
+                logger.info("Creating new preference file " + prefsFilePath);
             }
             initializedPrefs = prefsOptional.orElse(new UserPrefs());
         } catch (DataConversionException e) {
-            logger.warning("UserPrefs file at " + prefsFilePath + " is not in the correct format. "
-                    + "Using default user prefs");
+            logger.warning(
+                    "Preference file at " + prefsFilePath
+                            + " might be missing or corrupted. Using default preferences");
```
damithc: Let's not do any fancy line wrappings. Wrap only when the line is close to the max line length allowed.

damithc (changes requested):
Mostly OK. Some nits suggested.

(a) The first commit is about changing the logging pattern, rather than purely adding log messages. Something like this?
```
Make file-creation logs not resemble errors
...............
At the first run, any missing data/config files are created
with a 'file not found' log message, which might alarm the user
as it looks like an error message. 

Let's tweak the log messages to not resemble error messages.
```
(b) How about this for the second commit? Feel free to enhance further.
Tip: ChatGPT is good at helping with text (.e.g., shorten, refine, fix grammar etc.)
```
Show file location in log messages
.........................
Let's include the file location in file-related log messages,
to make them more informative.
```
(c) Have you checked the whole code base to see if there are other log messages that can benefit from a similar tweak?

Eclipse-Dominator:
I have modified and standardized some logs. Regarding the use of full stop, adding full stop to certain messagaes in the format like `"TXT" + filename` can feel rather messy, I have standardised a few log messages in the MainApp.java, I think subsequent standardisation should be done via a separate issue since this issue to rephrase file not found errors to be less error-like

damithc:
>I have standardised a few log messages in the MainApp.java, I think subsequent standardisation should be done via a separate issue since this issue to rephrase file not found errors to be less error-like

@Eclipse-Dominator Yes, this PR should not touch other log messages. But the log messages touched in this PR should be _locally consistent_ (i.e., consistent with the surrounding code).

src/main/java/seedu/address/MainApp.java
```diff
@@ -177,7 +187,7 @@ public void stop() {
         try {
             storage.saveUserPrefs(model.getUserPrefs());
         } catch (IOException e) {
-            logger.severe("Failed to save preferences " + StringUtil.getDetails(e));
+            logger.severe("Failed to save preferences : " + StringUtil.getDetails(e));
```
damithc: Unrelated? Can add it as a separate commit if the value added is significant. Usually, we don't use a space before `:`.

src/main/java/seedu/address/MainApp.java
```diff
@@ -118,12 +121,12 @@ protected Config initConfig(Path configFilePath) {
         try {
             Optional<Config> configOptional = ConfigUtil.readConfig(configFilePathUsed);
             if (!configOptional.isPresent()) {
-                logger.info("Config file not found. Will be starting with a new default config file.");
+                logger.info("Creating new config file " + configFilePathUsed);
```
damithc, line 124: Should this change be split between the two commits?

damithc (changes requested):
Sorry, I missed this earlier. The second commit does more than add file location. We can tweak the commit message accordingly. Something like this?

```
Make file-related log messages more informative
....
Let's make file-related log messages more informative, as follows:
* include the file location when referring to a file
* use a more informative term: prefs -> preferences
```

src/main/java/seedu/address/MainApp.java
Eclipse-Dominator: I was following the other instances where colon is used in the commit as I noticed all of them included an additional space before the colon.

src/main/java/seedu/address/MainApp.java
Eclipse-Dominator: I don't think the first commit originally modified the line that's why it was added to the 2nd commit. 
If we were to split, is it something like putting "Creating new config file." in the first commit 
and 
the 2nd commit to adding the file directory?

damithc: Yes, split like that, to be consistent with the rest. I guess it was missed the first time round.

src/main/java/seedu/address/MainApp.java
damithc: Then it should be fine. Local consistency matters.

src/main/java/seedu/address/MainApp.java
```diff
@@ -145,6 +148,9 @@ protected UserPrefs initPrefs(UserPrefsStorage storage) {
         UserPrefs initializedPrefs;
         try {
             Optional<UserPrefs> prefsOptional = storage.readUserPrefs();
+            if (!prefsOptional.isPresent()) {
+                logger.info("Prefs file not found. Will be starting with a new default prefs file.");
```
damithc: Why the phrasing of this is different from the config file message?

src/main/java/seedu/address/MainApp.java
damithc: This change does not belong in this commit.

damithc (changes requested):
Some further nits w.r.t., what goes in which commit.

src/main/java/seedu/address/MainApp.java
Eclipse-Dominator: My bad, I think I might have messed up during the rebasing, thanks for pointing it out. Really need to be more careful in the future regarding this.

src/main/java/seedu/address/MainApp.java
```diff
@@ -143,21 +146,22 @@ protected Config initConfig(Path configFilePath) {
...
             if (!prefsOptional.isPresent()) {
-                logger.info("Creating new prefs file.");
+                logger.info("Creating new preference file " + prefsFilePath);
             }
             initializedPrefs = prefsOptional.orElse(new UserPrefs());
         } catch (DataConversionException e) {
-            logger.warning("UserPrefs file at " + prefsFilePath + " is not in the correct format. "
-                    + "Using default user prefs");
+            logger.warning("Preference file at " + prefsFilePath
+                    + " might be missing or corrupted. Using default preferences.");
```
damithc: Any reason for the change of the log message text here?

damithc (commented):
Just a query only.

damithc:
@se-edu/tech-team-level1 for your review, if any of you are available these days. 
Please review each commit independently, including the commit message. Ensure the diff matches the commit message. Don't hesitate to nitpick, as we want to polish up each commit as much as we can, before merging.

src/main/java/seedu/address/MainApp.java
Eclipse-Dominator: In the previous PR, #152, the "Prefs file" is used to describe the preference file. Thus I wanted to standardize it to the same term and replaced UserPrefs with Preference instead.

damithc: Sorry, I should have been more specific. Why go from 'not in the correct format' to 'might be missing or corrupted'?

Eclipse-Dominator: I have standardized all corrupted warnings to use " is not in the correct format."

damithc: In this case it is more about giving the accurate error message rather than standardizing. Which one is more accurate?

Eclipse-Dominator: Not in the correct format/corrupted will be more accurate here. Since missing file will be reflected by "Creating XXX file." log

damithc: I see. I guess we can keep it as 'not in the correct format' (the incorrect format may be caused by many reasons, including file corruption).

src/main/java/seedu/address/MainApp.java
```diff
@@ -118,12 +121,12 @@ protected Config initConfig(Path configFilePath) {
...
             if (!configOptional.isPresent()) {
-                logger.info("Creating new config file.");
+                logger.info("Creating new config file " + configFilePathUsed);
             }
             initializedConfig = configOptional.orElse(new Config());
         } catch (DataConversionException e) {
-            logger.warning("Config file at " + configFilePathUsed + " is not in the correct format. "
-                    + "Using default config properties");
+            logger.warning("Config file at " + configFilePathUsed
+                    + " is not in the correct format. Using default config properties.");
```
damithc: Previous line wrapping position is better?

src/main/java/seedu/address/MainApp.java
```diff
@@ -143,21 +146,22 @@ protected Config initConfig(Path configFilePath) {
...
             if (!prefsOptional.isPresent()) {
-                logger.info("Creating new prefs file.");
+                logger.info("Creating new preference file " + prefsFilePath);
             }
             initializedPrefs = prefsOptional.orElse(new UserPrefs());
         } catch (DataConversionException e) {
-            logger.warning("UserPrefs file at " + prefsFilePath + " is not in the correct format. "
-                    + "Using default user prefs");
+            logger.warning("Preference file at " + prefsFilePath
+                    + " is not in the correct format. Using default preferences.");
```
damithc: Same here

damithc (commented):
Just a very minor line wrapping issue. Before posting a new iteration, we can wait 1-2 days to see if there are any more reviews from others.

src/main/java/seedu/address/MainApp.java
Eclipse-Dominator: I think it's around the same since the new one joins the 2nd half of the string together. 
But the older formatting does seem to improve readability.

damithc: For consistency, wrap when you are close to the max length. Don't wrap in arbitrary places for the sake of readability unless there is a strong justification. 
Reason: While readability is good, line wrapping differently to improve readability would cause us to make case-by-case decisions every time we do line wrapping. It's not worth the effort.

Eclipse-Dominator:
Updated iteration fixing the line breaks, please update if there is any more changes required @se-edu/tech-team-level1

PR #172, opened 2023-05-11
[#73] no write access fail gracefully

### LINKED ISSUE #73: Fail gracefully if the application is not allowed to write files

Eclipse-Dominator:
Fixes #73 
Continued from #77 

When there is insufficient permission to write data to the files, an `AccessDeniedException` will appear in the error message. 
This is not as informative as the error message is the same for both `AccessDeniedException` and `IOException`. Thus let's make the error message more descriptive for both exceptions.

Besides this, the same error message (`Invalid Command: <command text>` is logged whenever a CommandException or ParseException occurs. However, such errors can be due to IOExceptions which is not related to the invalid command. 

Therefore, there is also a need to improve the accuracy of the error message by changing it to more accurately describe the failure.

### DIFF
--- a/src/main/java/seedu/address/logic/LogicManager.java
+++ b/src/main/java/seedu/address/logic/LogicManager.java
@@ -1,6 +1,7 @@
 package seedu.address.logic;
 
 import java.io.IOException;
+import java.nio.file.AccessDeniedException;
 import java.nio.file.Path;
 import java.util.logging.Logger;
 
@@ -21,7 +22,11 @@ import seedu.address.storage.Storage;
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_FORMAT = "Could not save data due to the following error: %s";
+
+    public static final String FILE_OPS_PERMISSION_ERROR_FORMAT =
+            "Could not save data to file %s due to insufficient permissions to write to the file or the folder.";
+
     private final Logger logger = LogsCenter.getLogger(LogicManager.class);
 
     private final Model model;
@@ -47,8 +52,10 @@ public class LogicManager implements Logic {
 
         try {
             storage.saveAddressBook(model.getAddressBook());
+        } catch (AccessDeniedException e) {
+            throw new CommandException(String.format(FILE_OPS_PERMISSION_ERROR_FORMAT, e.getMessage()), e);
         } catch (IOException ioe) {
-            throw new CommandException(FILE_OPS_ERROR_MESSAGE + ioe, ioe);
+            throw new CommandException(String.format(FILE_OPS_ERROR_FORMAT, ioe.getMessage()), ioe);
         }
 
         return commandResult;
--- a/src/main/java/seedu/address/ui/MainWindow.java
+++ b/src/main/java/seedu/address/ui/MainWindow.java
@@ -188,7 +188,7 @@ public class MainWindow extends UiPart<Stage> {
 
             return commandResult;
         } catch (CommandException | ParseException e) {
-            logger.info("Invalid command: " + commandText);
+            logger.info("An error occurred while executing command: " + commandText);
             resultDisplay.setFeedbackToUser(e.getMessage());
             throw e;
         }
--- a/src/test/java/seedu/address/logic/LogicManagerTest.java
+++ b/src/test/java/seedu/address/logic/LogicManagerTest.java
@@ -11,6 +11,7 @@ import static seedu.address.testutil.Assert.assertThrows;
 import static seedu.address.testutil.TypicalPersons.AMY;
 
 import java.io.IOException;
+import java.nio.file.AccessDeniedException;
 import java.nio.file.Path;
 
 import org.junit.jupiter.api.BeforeEach;
@@ -33,7 +34,8 @@ import seedu.address.storage.StorageManager;
 import seedu.address.testutil.PersonBuilder;
 
 public class LogicManagerTest {
-    private static final IOException DUMMY_IO_EXCEPTION = new IOException("dummy exception");
+    private static final IOException DUMMY_IO_EXCEPTION = new IOException("dummy IO exception");
+    private static final IOException DUMMY_AD_EXCEPTION = new AccessDeniedException("dummy access denied exception");
 
     @TempDir
     public Path temporaryFolder;
@@ -70,22 +72,14 @@ public class LogicManagerTest {
 
     @Test
     public void execute_storageThrowsIoException_throwsCommandException() {
-        // Setup LogicManager with JsonAddressBookIoExceptionThrowingStub
-        JsonAddressBookStorage addressBookStorage =
-                new JsonAddressBookIoExceptionThrowingStub(temporaryFolder.resolve("ioExceptionAddressBook.json"));
-        JsonUserPrefsStorage userPrefsStorage =
-                new JsonUserPrefsStorage(temporaryFolder.resolve("ioExceptionUserPrefs.json"));
-        StorageManager storage = new StorageManager(addressBookStorage, userPrefsStorage);
-        logic = new LogicManager(model, storage);
+        assertCommandFailureForExceptionFromStorage(DUMMY_IO_EXCEPTION, String.format(
+                LogicManager.FILE_OPS_ERROR_FORMAT, DUMMY_IO_EXCEPTION.getMessage()));
+    }
 
-        // Execute add command
-        String addCommand = AddCommand.COMMAND_WORD + NAME_DESC_AMY + PHONE_DESC_AMY + EMAIL_DESC_AMY
-                + ADDRESS_DESC_AMY;
-        Person expectedPerson = new PersonBuilder(AMY).withTags().build();
-        ModelManager expectedModel = new ModelManager();
-        expectedModel.addPerson(expectedPerson);
-        String expectedMessage = LogicManager.FILE_OPS_ERROR_MESSAGE + DUMMY_IO_EXCEPTION;
-        assertCommandFailure(addCommand, CommandException.class, expectedMessage, expectedModel);
+    @Test
+    public void execute_storageThrowsAdException_throwsCommandException() {
+        assertCommandFailureForExceptionFromStorage(DUMMY_AD_EXCEPTION, String.format(
+                LogicManager.FILE_OPS_PERMISSION_ERROR_FORMAT, DUMMY_AD_EXCEPTION.getMessage()));
     }
 
     @Test
@@ -147,16 +141,35 @@ public class LogicManagerTest {
     }
 
     /**
-     * A stub class to throw an {@code IOException} when the save method is called.
+     * Tests the Logic component's handling of an {@code IOException} thrown by the Storage component.
+     *
+     * @param e the exception to be thrown by the Storage component
+     * @param expectedMessage the message expected inside exception thrown by the Logic component
      */
-    private static class JsonAddressBookIoExceptionThrowingStub extends JsonAddressBookStorage {
-        private JsonAddressBookIoExceptionThrowingStub(Path filePath) {
-            super(filePath);
-        }
+    private void assertCommandFailureForExceptionFromStorage(IOException e, String expectedMessage) {
+        Path prefPath = temporaryFolder.resolve("ExceptionUserPrefs.json");
 
-        @Override
-        public void saveAddressBook(ReadOnlyAddressBook addressBook, Path filePath) throws IOException {
-            throw DUMMY_IO_EXCEPTION;
-        }
+        // Inject LogicManager with an AddressBookStorage that throws the IOException e when saving
+        JsonAddressBookStorage addressBookStorage = new JsonAddressBookStorage(prefPath) {
+            @Override
+            public void saveAddressBook(ReadOnlyAddressBook addressBook, Path filePath)
+                    throws IOException {
+                throw e;
+            }
+        };
+
+        JsonUserPrefsStorage userPrefsStorage =
+                new JsonUserPrefsStorage(temporaryFolder.resolve("ExceptionUserPrefs.json"));
+        StorageManager storage = new StorageManager(addressBookStorage, userPrefsStorage);
+
+        logic = new LogicManager(model, storage);
+
+        // Triggers the saveAddressBook method by executing an add command
+        String addCommand = AddCommand.COMMAND_WORD + NAME_DESC_AMY + PHONE_DESC_AMY
+                + EMAIL_DESC_AMY + ADDRESS_DESC_AMY;
+        Person expectedPerson = new PersonBuilder(AMY).withTags().build();
+        ModelManager expectedModel = new ModelManager();
+        expectedModel.addPerson(expectedPerson);
+        assertCommandFailure(addCommand, CommandException.class, expectedMessage, expectedModel);
     }
 }


Eclipse-Dominator:
Apologies for all the changes, I realised I forgot to include a line spacing between the commit header and the commit body

damithc:
> There is an issue with the exit command where exit will attempt to save all file before exiting and due to the permission error, it will prevent the program from exiting.

hmm... there is no real need to save data upon exit. Perhaps we can skip the data saving if the command is the exit command?

Eclipse-Dominator:
Hmm I don't think that there is a very clean way to identify specific command without hardcoding certain checks. 

```java
@Override
    public CommandResult execute(String commandText) throws CommandException, ParseException {
        logger.info("----------------[USER COMMAND][" + commandText + "]");

        CommandResult commandResult;
        Command command = addressBookParser.parseCommand(commandText);
        commandResult = command.execute(model);

        try {
            storage.saveAddressBook(model.getAddressBook());
        } catch (IOException ioe) {
            throw new CommandException(FILE_OPS_ERROR_MESSAGE + ioe, ioe);
        }

        return commandResult;
    }
```

```java
    private CommandResult executeCommand(String commandText) throws CommandException, ParseException {
        try {
            CommandResult commandResult = logic.execute(commandText);
            logger.info("Result: " + commandResult.getFeedbackToUser());
            resultDisplay.setFeedbackToUser(commandResult.getFeedbackToUser());

            if (commandResult.isShowHelp()) {
                handleHelp();
            }

            if (commandResult.isExit()) {
                handleExit();
            }

            return commandResult;
        } catch (CommandException | ParseException e) {
            logger.info("Invalid command: " + commandText);
            resultDisplay.setFeedbackToUser(e.getMessage());
            throw e;
        }
    }
```
Saving of addressbook is done in execute itself and only afterwards verification of whether something is help or exit is done. 

---
I was thinking of allowing execute to throw a custom exception to represent critical error where the program need to exit instead of a CommandException
```java
        try {
            storage.saveAddressBook(model.getAddressBook());
        } catch (IOException ioe) {
            throw new CriticalException(FILE_OPS_ERROR_MESSAGE + ioe, ioe); // <-- 
        }
```
Then specifically catching it in `MainWindow.java`
```java
        catch (CriticalException e) {
            handleCriticalError(e); // <-- spawns a popup that exits program after closing. 
        }
```
---
Alternatively, we can also store the exception in CommandResult instead of catching it later on. In a sense we are decoupling the save exception from the parse exceptions. However, both of these method could be an overkill to handle permission denied error.

damithc:
As this is a very rare case, we should try a solution that minimizes changes to the code. So, approach 3 should be fine.
Do you think this PR can be just one commit?

Eclipse-Dominator:
I think 2 commit is probably more suited here?
1. Decoupling IOException and AccessDeniedException
2. Decoupling ParseException and CommandException

As for the commit body, I am thinking of 

1.  
```
Better error message when file have no write-access

When the data file has no write access, a message containing.
a raw exception is displayed to the user with no specific instructions
on handling such errors.

Let's 
* Polish up error messages displayed to the user
* Decouple IOExceptions and AccessDeniedExceptions
* Add instructions to ask the users to manually close AB3
```
2. 
```
Decoupling of ParseException and CommandException log messages

All errors related to command is logged as "invalid command".
Let's be more specific by decoupling ParseExceptions and CommandExceptions
to enable an more accurate logging of the errors.
```

damithc:
I'm not sure if the second commit adds enough value. While the same prefix is used in both cases, the error message itself should tell us what exactly happened, right?
Don't forget that commit message subject should be in imperative mood.

Eclipse-Dominator:
I have updated the commit to the following:
```
Improve error message if no write permission

Insufficient permission to read the data file will show
an AccessDeniedException as error message.
AB3 should be able to fail gracefully with instructions
to close the application upon failure.

Let's
- make the error message more user friendly by informing
the user that there are insufficient permissions for the file.
- Decouple AccessDeniedException from IOException messages.
- Decouple ParseException and CommandException log messages.
```

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,11 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: %s\n"
+            + "Please exit the application immediately via the exit button";
```
damithc: Is there a better alternative to 'via the exit button'? The reader might wonder "what's the 'exit button'?"

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -47,8 +52,12 @@ public CommandResult execute(String commandText) throws CommandException, ParseE
 
         try {
             storage.saveAddressBook(model.getAddressBook());
+        } catch (AccessDeniedException e) {
+            String message = e.getMessage();
```
damithc: inline this, as it is used only once?

src/main/java/seedu/address/logic/LogicManager.java
damithc: Exiting doesn't solve the problem. We can inform user how to work around the error.

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -187,10 +187,14 @@ private CommandResult executeCommand(String commandText) throws CommandException
...
-        } catch (CommandException | ParseException e) {
+        } catch (ParseException e) {
             logger.info("Invalid command: " + commandText);
             resultDisplay.setFeedbackToUser(e.getMessage());
             throw e;
+        } catch (CommandException e) {
+            logger.info("An error occured while executing this command: " + commandText);
+            resultDisplay.setFeedbackToUser(e.getMessage());
+            throw e;
         }
```
damithc, line 194: I don't think this change is related to the issue at hand. It should be a separate PR.
Also, I think the following tweak is enough (no need for two catch blocks).
```suggestion
        } catch (CommandException | ParseException e) {
            logger.info("An error occured while executing this command: " + commandText);
            resultDisplay.setFeedbackToUser(e.getMessage());
            throw e;
        }
```

damithc (changes requested):
Some suggestions added.

Commit message: Use 'application' instead of 'AB3' as other higher AB versions (e.g., AB4) are derived from this code base as well.

src/main/java/seedu/address/logic/LogicManager.java
Eclipse-Dominator: Maybe we can use 'X' button? I don't think there is a proper term that is easily understandable for the exit button.

I have included the following "work arounds":
            "1. Running the application as an administrator.\n"
            "2. Changing the permissions of the save files or folder to allow read write access.\n"
            "3. Changing the save location of the application to a location that you have read write access to.\n";

src/main/java/seedu/address/ui/MainWindow.java
Eclipse-Dominator: I separated the 2 because in the log file, any command that triggered CommandException or ParseException previously will log the line "Invalid Command: %s", even though the command is valid.

Perhaps we can standardise to a general term "An error occured while executing this command: " instead of invalid command as suggested?

damithc: >any command that triggered CommandException or ParseException previously will log the line "Invalid Command: %s", even though the command is valid.

I agree, that doesn't sound right.

> Perhaps we can standardise to a general term "An error occured while executing this command: " instead of invalid command as suggested?

Yes, that's what I'm suggesting as well.

src/main/java/seedu/address/logic/LogicManager.java
damithc: > Maybe we can use 'X' button? I don't think there is a proper term that is easily understandable for the exit button.

Yes that might work.  e.g., `exit the application by clicking the eXit button (i.e., the button marked with an 'X' icon)`


>3\. Changing the save location of the application to a location that you have read write access to.

The user might wonder how to change the save location.

Perhaps we can just tell the user the _intent_ of the workaround and let her figure out a mechanism? e.g., `Before running it again, ensure the application .jar file has write permissions to the folder it is in.`

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,13 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: %s\n"
+            + "Please exit the application immediately via the exit button.";
```
damithc: Probably no need for the word `immediately`?

damithc: `via` -> `by clicking`

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,13 @@
...
+    public static final String FILE_OPS_PERM_ERROR_MESSAGE = "Could not save data to file: "
```
damithc: `PERM` is not a standard abbreviation. Use the full word.

src/main/java/seedu/address/logic/LogicManager.java
damithc: Why does this message asks the user to exit the application but doesn't say how to recover?

damithc (changes requested):
Commit message: 
* subject says `write`, first sentence says `read`
* what we are doing is making the error message more informative (rather than 'fail gracefully), right? If so, we can omit the phrase 'fail gracefully'.
* I don't think we are doing any intentional 'decoupling' in this PR. It's just incidental that we add a new exception branch.

src/main/java/seedu/address/logic/LogicManager.java
Eclipse-Dominator: This error message is for generic IOException errors that are not AccessDeniedExceptions, so I am not sure what the standard procedure is for recovering from such errors, so I left it as such.

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,13 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: %s\n"
+            + "Please exit the application by clicking the exit button (i.e., the button marked with an 'X' icon).";
```
damithc: I still think either you have to tell the user how to recover, or just state the problem only (i.e., without mention exiting at all).

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,13 @@
...
+    public static final String FILE_OPS_PERMISSION_ERROR_MESSAGE = "Could not save data to file: "
+            + "Insufficient Permissions for %s.\n\n"
+            + "Please exit the application by clicking the exit button (i.e., the button marked with an 'X' icon).\n"
```
damithc: It's been a while so I might have forgotten the reason; why the user has to exit the application?

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,13 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: %s\n"
+            + "Please exit the application by clicking the exit button (i.e., the button marked with an 'X' icon).";
+    public static final String FILE_OPS_PERMISSION_ERROR_MESSAGE = "Could not save data to file: "
+            + "Insufficient Permissions for %s.\n\n"
```
damithc: Permissions -> permissions?

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -84,7 +86,28 @@ public void execute_storageThrowsIoException_throwsCommandException() {
...
+        // Execute add command
+        String addCommand =
+                AddCommand.COMMAND_WORD + NAME_DESC_AMY + PHONE_DESC_AMY + EMAIL_DESC_AMY + ADDRESS_DESC_AMY;
+        Person expectedPerson = new PersonBuilder(AMY).withTags().build();
+        ModelManager expectedModel = new ModelManager();
+        expectedModel.addPerson(expectedPerson);
+        String expectedMessage =
+                String.format(LogicManager.FILE_OPS_PERMISSION_ERROR_MESSAGE, DUMMY_AD_EXCEPTION.getMessage());
         assertCommandFailure(addCommand, CommandException.class, expectedMessage, expectedModel);
     }
```
damithc, line 83: Good job adding a test. 
Looks almost identical to an existing test method. Perhaps the common code can be extracted out as a utility method?

damithc (changes requested):
Some comments on a few things that I didn't think of before.

src/main/java/seedu/address/logic/LogicManager.java
Eclipse-Dominator: Not exiting the application is fine but none of the changes will be saved until the permissions is changed. Perhaps we can instruct the user to change the permissions of the relevant files/folder on the go so that future commands will be saved but there could be complications so it might be better to just exit the application straight away and change the permission rather than changing the permissions on the go

Exiting via the x button is required as typing exit will not cause the application to quit

src/main/java/seedu/address/logic/LogicManager.java
Eclipse-Dominator: I think I'll state the error and mention continued usage might lead to data loss

src/main/java/seedu/address/logic/LogicManager.java
Eclipse-Dominator: I have updated the message to 
```
Could not save data to file: Insufficient permissions for FILENAME

Please ensure the application .jar file has write permissions to the folder it is in.
Continued usage of the application without fixing this error might lead to data loss.
You can exit the application by clicking the exit button (i.e., the button marked with an 'X' icon).
```

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -147,16 +163,19 @@ private void assertCommandFailure(String inputCommand, Class<? extends Throwable
...
 
     /**
-     * A stub class to throw an {@code IOException} when the save method is called.
+     * A method that returns a {@code StorageManager} with the {@code JsonAddressBookStorage} that
```
damithc: This is not how we phrase the overview statement of methods.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -147,16 +163,19 @@ private void assertCommandFailure(String inputCommand, Class<? extends Throwable
...
+        // Setup LogicManager with JsonAddressBookAcessDemoedExceptionThrowingStub
+        JsonAddressBookStorage addressBookStorage = new JsonAddressBookStorage(temporaryFolder) {
+            @Override
+            public void saveAddressBook(ReadOnlyAddressBook addressBook, Path filePath) throws IOException {
+                throw exception;
+            }
+        };
+        JsonUserPrefsStorage userPrefsStorage =
+                new JsonUserPrefsStorage(temporaryFolder.resolve("ExceptionUserPrefs.json"));
+        return new StorageManager(addressBookStorage, userPrefsStorage);
```
damithc: Seems not worth extracting this. The extra long name is also an indication of extra complexity -- the v5 stub class is easier to understand than this stub-creating-method. 

How about something like this (because the only difference between the two test cases is the exception and the expected error message)?
```java
    @Test
    public void execute_storageThrowsIoException_throwsCommandException() {
        assertCommandFailureForException(DUMMY_AD_EXCEPTION, 
            String.format(LogicManager.FILE_OPS_PERMISSION_ERROR_MESSAGE, DUMMY_AD_EXCEPTION.getMessage());
    }
    
    @Test
    public void execute_storageThrowsIoException_throwsCommandException() {
        assertCommandFailureForException(DUMMY_IO_EXCEPTION,
            String.format(LogicManager.FILE_OPS_ERROR_MESSAGE, DUMMY_IO_EXCEPTION.getMessage()));
    }
```

Alternatively, we can go back to the v5 test cases with the duplicated code. We can generalize further when we have a third similar test case.

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,14 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data! The following errors has occured: %s\n"
+            + "Continued usage of the application without fixing this error might lead to data loss.\n";
```
damithc: Let's remove this line as well. Reasons: 1. it seems like an obvious thing. 2. it's talking about _consequence_ of the error from an overall application's perspective. It's cleaner to stick with just the description of the error only.

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,14 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data! The following errors has occured: %s\n"
```
damithc: Unless this error is more severe than the other one, no good reason to use an `!` but not the other.  The phrasing of the error message can be same (for consistency) except there is a good reason to differ.

damithc (changes requested):

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,12 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data. The following errors has occured: %s\n";
+
+    public static final String FILE_OPS_PERMISSION_ERROR_MESSAGE = "Could not save data to file: "
+            + "Insufficient permissions to write content to %s.\n\n"
+            + "The application currently do not have write permissions to the folder it is in.\n";
```
damithc: Are we able to say this for sure? What if the file exists but is read-only? 
Also, the fact that the data file is created in the home folder a separate piece of information that might change independent of this code. So, I don't think it is safe to say something like this here. What do you think?

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -147,16 +141,38 @@ private void assertCommandFailure(String inputCommand, Class<? extends Throwable
...
 
     /**
-     * A stub class to throw an {@code IOException} when the save method is called.
+     * Execute a dummy command in a {@code LogicManager} containg a {@code Storage} Stub that will
+     * throw an {@code IOException}, this will confirm that <br>
+     * - the {@code expectedException} is thrown <br>
+     * - the resulting error message is equal to {@code expectedMessage} <br>
+     *
+     * @param e exception to be thrown by the AddressBookStorage
+     * @param expectedMessage expected message to be thrown
```
damithc: Execute -> Executes

Explain WHAT the method does rather than HOW it does it.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -147,16 +141,38 @@ private void assertCommandFailure(String inputCommand, Class<? extends Throwable
...
+        Path prefPath = temporaryFolder.resolve("ExceptionUserPrefs.json");
+
+        // Setup LogicManager with an AddressBookStorage that throws an IOException when saving
+        JsonAddressBookStorage addressBookStorage = new JsonAddressBookStorage(prefPath) {
+            @Override
+            public void saveAddressBook(ReadOnlyAddressBook addressBook, Path filePath)
+                    throws IOException {
+                throw e;
+            }
+        };
```
damithc, line 159: Nice way to avoid an extra class.  👍

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -147,16 +141,38 @@ private void assertCommandFailure(String inputCommand, Class<? extends Throwable
...
-    private static class JsonAddressBookIoExceptionThrowingStub extends JsonAddressBookStorage {
-        private JsonAddressBookIoExceptionThrowingStub(Path filePath) {
-            super(filePath);
-        }
-
-        @Override
-        public void saveAddressBook(ReadOnlyAddressBook addressBook, Path filePath) throws IOException {
-            throw DUMMY_IO_EXCEPTION;
-        }
+    private void assertCommandFailureForException(IOException e, String expectedMessage) {
```
damithc: Make the name a bit more specific?
```suggestion
    private void assertCommandFailureForExceptionFromStorage(IOException e, String expectedMessage) {
```

damithc (changes requested):
Some nits ...

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,11 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data. The following errors has occured: %s\n";
+
+    public static final String FILE_OPS_PERMISSION_ERROR_MESSAGE = "Could not save data to file: "
+            + "Insufficient permissions to write content to file %s.\n";
```
damithc:
```suggestion
    public static final String FILE_OPS_PERMISSION_ERROR_MESSAGE = "Could not save data to file. "
            + "Insufficient permissions to write to the file %s.\n";
```
* Made punctuation more consistent.
* Removed redundant word `content`

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -33,7 +34,8 @@
 import seedu.address.testutil.PersonBuilder;
 
 public class LogicManagerTest {
-    private static final IOException DUMMY_IO_EXCEPTION = new IOException("dummy exception");
+    private static final IOException DUMMY_IO_EXCEPTION = new IOException("dummy io exception");
+    private static final IOException DUMMY_AD_EXCEPTION = new AccessDeniedException("dummy ad exception");
```
damithc:
```suggestion
    private static final IOException DUMMY_IO_EXCEPTION = new IOException("dummy IO exception");
    private static final IOException DUMMY_AD_EXCEPTION = new AccessDeniedException("dummy access denied exception");
```

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -70,22 +72,14 @@ public void execute_validCommand_success() throws Exception {
...
-                + ADDRESS_DESC_AMY;
-        Person expectedPerson = new PersonBuilder(AMY).withTags().build();
-        ModelManager expectedModel = new ModelManager();
-        expectedModel.addPerson(expectedPerson);
-        String expectedMessage = LogicManager.FILE_OPS_ERROR_MESSAGE + DUMMY_IO_EXCEPTION;
-        assertCommandFailure(addCommand, CommandException.class, expectedMessage, expectedModel);
+    @Test
+    public void execute_storageThrowsAdException_throwsCommandException() {
+        assertCommandFailureForExceptionFromStorage(DUMMY_AD_EXCEPTION, String.format(
+                LogicManager.FILE_OPS_PERMISSION_ERROR_MESSAGE, DUMMY_AD_EXCEPTION.getMessage()));
```
damithc: Inconsistent line wrapping compared to line 75 (`String` vs `String.format(`)

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -147,16 +141,38 @@ private void assertCommandFailure(String inputCommand, Class<? extends Throwable
     }
 
     /**
-     * A stub class to throw an {@code IOException} when the save method is called.
+     * Test the handling of an {@code IOException} in a dummy command execution scenario by the
+     * Storage component. This test verifies that:<br>
+     * - The expected exception ({@code e}) is thrown<br>
+     * - The resulting error message matches the expected message ({@code expectedMessage})<br>
```
damithc:
```suggestion
     * Tests the Logic component's handling of an {@code IOException} thrown by 
     *  the Storage component. 
```

* Test -> Tests
* Shortened the description, as test descriptions need not be very detailed. The previous details were not complete anyway, as it did not mention about verifying the Model component is as expected.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -147,16 +141,38 @@ private void assertCommandFailure(String inputCommand, Class<? extends Throwable
...
+     *
+     * @param e the exception to be thrown by the AddressBookStorage
+     * @param expectedMessage the expected message to be thrown
```
damithc:
```suggestion
     * @param e the exception to be thrown by the Storage component 
     * @param expectedMessage the message expected inside exception thrown by the Logic component
```

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -147,16 +141,38 @@ private void assertCommandFailure(String inputCommand, Class<? extends Throwable
...
-        }
-
-        @Override
-        public void saveAddressBook(ReadOnlyAddressBook addressBook, Path filePath) throws IOException {
-            throw DUMMY_IO_EXCEPTION;
-        }
+    private void assertCommandFailureForExceptionFromStorage(IOException e, String expectedMessage) {
+        Path prefPath = temporaryFolder.resolve("ExceptionUserPrefs.json");
+
+        // Setup LogicManager with an AddressBookStorage that throws an IOException when saving
```
damithc:
```suggestion
        // Inject LogicManager with an AddressBookStorage that throws the IOException e when saving
```

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -147,16 +141,38 @@ private void assertCommandFailure(String inputCommand, Class<? extends Throwable
...
+            }
+        };
+
+        JsonUserPrefsStorage userPrefsStorage =
+                new JsonUserPrefsStorage(temporaryFolder.resolve("ExceptionUserPrefs.json"));
+        StorageManager storage = new StorageManager(addressBookStorage, userPrefsStorage);
+
+        logic = new LogicManager(model, storage);
+
+        // Execute add command
```
damithc:
```suggestion
        // To trigger the save method, execute an add command
```

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -188,7 +188,7 @@ private CommandResult executeCommand(String commandText) throws CommandException
 
             return commandResult;
         } catch (CommandException | ParseException e) {
-            logger.info("Invalid command: " + commandText);
+            logger.info("An error occured while executing this command: " + commandText);
```
damithc: Should this change be in a separate commit, as it is not related to other changes in this commit?

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -47,8 +52,10 @@ public CommandResult execute(String commandText) throws CommandException, ParseE
 
         try {
             storage.saveAddressBook(model.getAddressBook());
+        } catch (AccessDeniedException e) {
+            throw new CommandException(String.format(FILE_OPS_PERMISSION_ERROR_MESSAGE, e.getMessage()), e);
         } catch (IOException ioe) {
-            throw new CommandException(FILE_OPS_ERROR_MESSAGE + ioe, ioe);
+            throw new CommandException(String.format(FILE_OPS_ERROR_MESSAGE, ioe.getMessage()), ioe);
         }
```
damithc, line 59: @Eclipse-Dominator What kind of problems would cause IOException to be thrown instead of AccessDeniedException ?

damithc (changes requested):
Added some stylistic suggestions. Feel free to deviate if you think of better alternatives.

Might need to change the commit message to reflect the current version of the code too.

src/main/java/seedu/address/logic/LogicManager.java
Eclipse-Dominator: There are some exceptions like DiskFullException that could be thrown. Generally it is part of a niche and unlikely scenarios that will cause an IOException to be thrown so this catch block is generally used to handle them.

damithc: @Eclipse-Dominator reminder about this question

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,11 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data. The following errors has occured: %s\n";
+
+    public static final String FILE_OPS_PERMISSION_ERROR_MESSAGE =
+            "Could not save data to file. Insufficient permissions to write to the file %s.\n";
```
damithc: `to the file (or the folder):`?

damithc (commented):
Commit 2: good. It's a tiny change but definite value added.
Commit 1: Feels like the value added is marginal (and code to value ratio is very low), but I'm OK to keep it. OK to drop it as well. The low value of the commit is not your fault -- it's because I changed my mind about giving recovery instructions to the user.

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,11 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data. The following errors has occured: %s\n";
+
+    public static final String FILE_OPS_PERMISSION_ERROR_MESSAGE =
+            "Could not save data to file (or the folder). Insufficient permissions to write to the file %s.\n";
```
damithc:
```suggestion
    public static final String FILE_OPS_PERMISSION_ERROR_MESSAGE =
            "Could not save data to file %s. Insufficient permissions to write to the file (or the folder).\n";
```

damithc (changes requested):
Cosmetic suggestion.
Also, update the PR description to match the current state so that we can ask others to review this PR.

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,11 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data. The following errors has occured: %s\n";
+
+    public static final String FILE_OPS_PERMISSION_ERROR_MESSAGE =
+            "Could not save data to file %s. Insufficient permissions to write to the file (or the folder).\n";
+
```
damithc, line 29: Just noticed the `\n` at the end. Doesn't seem to belong there, as the error message is a one-liner.

damithc (commented):
Minor cosmetic issue.

damithc:

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,11 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data. The following errors has occured: %s\n";
```
chia-yh:
```suggestion
    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data. The following error has occured: %s.\n";
```

chia-yh: Not sure if the comment about `\n` at the end for `FILE_OPS_PERMISSION_ERROR_MESSAGE` also applies here

chia-yh (commented):
Some minor nits about the error message, but otherwise looks fine. I like the change to use `String.format` for the error message, as it feels cleaner. I also think it's an excellent idea to have the error message be more informative/accurate.

SPWwj (commented):
image

Just a minor sugguestion on the user interface. Currently, the exception messages are too long, causing some readability issues. It may be more user-friendly to insert line breaks into these messages to enhance their legibility.

Other than this, everything look good!

damithc:
> Just a minor sugguestion on the user interface. Currently, the exception messages are too long, causing some readability issues. It may be more user-friendly to insert line breaks into these messages to enhance their legibility.

Fair point @SPWwj Probably OK in this case because the second sentence follows directly from the first. We don't want to give the impression that it is a separate problem. In fact I wonder if we should phrase it as `...book.json due to insufficient ...` 🤔

SPWwj:
> > Just a minor sugguestion on the user interface. Currently, the exception messages are too long, causing some readability issues. It may be more user-friendly to insert line breaks into these messages to enhance their legibility.
> 
> Fair point @SPWwj Probably OK in this case because the second sentence follows directly from the first. We don't want to give the impression that it is a separate problem. In fact I wonder if we should phrase it as `...book.json due to insufficient ...` 🤔

Agreed, this completes the sentence, enhancing clarity and precision by providing a reason. In circumstances with multiple issues, it ensures that no ambiguity arises from multiple sentences.

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,11 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data. The following errors has occured: %s";
+
+    public static final String FILE_OPS_PERMISSION_ERROR_MESSAGE =
+            "Could not save data to file %s due to insufficient permissions to write to the file (or the folder).\n";
```
SPWwj:
```suggestion
            "Could not save data to file %s due to insufficient permissions to write to the file or the folder.";
```

1. My personal opinion is that components using the strings should handle additional formatting elements, such as newline characters. I invite further discussion on this.
2. In my view, the phrase "(or the folder)" is redundant. Although it is commonly seen, it doesn't offer any differentiation from "or the folder," which is two characters fewer and provides a cleaner presentation.

Eclipse-Dominator: AB3 right now does not really support composing messages, especially exceptions, the closest element that is related to this is `MainWindow` where the related method is shown below:

```java
/**
     * Executes the command and returns the result.
     *
     * @see seedu.address.logic.Logic#execute(String)
     */
    private CommandResult executeCommand(String commandText) throws CommandException, ParseException {
        try {
            CommandResult commandResult = logic.execute(commandText);
            logger.info("Result: " + commandResult.getFeedbackToUser());
            resultDisplay.setFeedbackToUser(commandResult.getFeedbackToUser());

            if (commandResult.isShowHelp()) {
                handleHelp();
            }

            if (commandResult.isExit()) {
                handleExit();
            }

            return commandResult;
        } catch (CommandException | ParseException e) {
            logger.info("An error occured while executing this command: " + commandText);
            resultDisplay.setFeedbackToUser(e.getMessage());
            throw e;
        }
    }
```

imo, the scenario right now is quite literally what you see is what you get for the handling of exceptions. For the `Exception`s  handled in this PR, the component handling and using the string is `LogicManager` So I think it is ok to remove the newline character and keep the handling as it is.


As for 2, I do agree that "(or the folder)" feels redundant but it is also possible to change the folder permission to readonly so it can be slightly different compared to the file. However, ultimately it is the file that does not have read permission due to the folder settings. 

I am okay with removing the phrase but I feel that adding it can be more accurate to the exception.

damithc: 1. I agree that `\n` should be removed. In fact I believe it was left there by mistake. 
2. I think @SPWwj is asking to remove the parenthesis, not the whole phrase.

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,11 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data. The following errors has occured: %s";
```
hansstanley: Just some minor grammar changes, assuming that only one error follows this message:

```suggestion
    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data. The following error has occurred: %s";
```

To ensure this format string is not used as-is instead of being wrapped in `String.format`, it might also be helpful to rename it to distinguish it from other errors messages, for e.g. `FILE_OPS_ERROR_TEMPLATE` or `FILE_OPS_ERROR_PREFIX`.

Eclipse-Dominator: Good catch, I will follow the formatting name of the classes here!

damithc:
@chia-yh @hansstanley @SPWwj Can we have another review, as the code has changed since your last review?

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,11 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_FORMAT = "Could not save data. The following error has occurred: %s";
```
SPWwj:
```suggestion
    public static final String FILE_OPS_ERROR_FORMAT = "Could not save data due to the following occurred error: %s.";
```

damithc: Perhaps remove `occurred` as well? i.e., `Could not save data due to the following error`

chia-yh (commented):
LGTM, though I also think the suggested changes to `FILE_OPS_ERROR_FORMAT` are a good idea as it reads better.

hansstanley (commented):
Other than the existing comments, LGTM

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -21,7 +22,11 @@
  * The main LogicManager of the app.
  */
 public class LogicManager implements Logic {
-    public static final String FILE_OPS_ERROR_MESSAGE = "Could not save data to file: ";
+    public static final String FILE_OPS_ERROR_FORMAT = "Could not save data due to the following error: %s";
+
+    public static final String FILE_OPS_PERMISSION_ERROR_FORMAT =
+            "Could not save data to file %s due to insufficient permissions to write to the file or the folder.";
```
SPWwj, line 28:
```suggestion
            "Could not perform write operations to file %s due to insufficient permissions";
```

src/main/java/seedu/address/ui/MainWindow.java
SPWwj, line 191: An error occured while executing command:

SPWwj:
```suggestion
            logger.info("An error occurred while executing command:: " + commandText);
```

src/main/java/seedu/address/logic/LogicManager.java
damithc: I wouldn't mind keeping the original, as it is more informative and context-specific.

src/main/java/seedu/address/ui/MainWindow.java
damithc: @SPWwj Should be only one colon?

SPWwj: Oh yes 😂

```suggestion
            logger.info("An error occurred while executing command: " + commandText);
```

src/main/java/seedu/address/logic/LogicManager.java
Eclipse-Dominator: I think the original feels more specific to the actual warning as well, `perform write operations` might be a bit too generic.

damithc (approved):
No further comments from me. LGTM. 👍

damithc:
Thanks for these improvements, @Eclipse-Dominator

PR #174, opened 2023-05-12
[#173] Null class logger not logging to logger output

### LINKED ISSUE #173: Null class logger not logging to logger output
When a logger is created with a null class, it's logs are not being logged into the log file when all logs should be recorded in the log file.

If we create a logger using a null class for instance 
```java
Logger logger = LogsCenter.getLogger(Object.class.getSuperclass());
// Object.class.getSuperclass() gives a null class value
// since getLogger is overload with string type, this prevent ambiguity when compiling.
```
Logs made via this logger will not be able to write logs to the addressbook.log.0 file. 

This is due to an incorrect function call to `LogsCenter.getLogger(str)` which enables logs to be written to a log file as well as other pipelines.

```java
    /**
     * Creates a logger with the given name.
     */
    public static Logger getLogger(String name) {
        Logger logger = Logger.getLogger(name);
        logger.setUseParentHandlers(false);

        removeHandlers(logger);
        addConsoleHandler(logger);
        addFileHandler(logger);

        return Logger.getLogger(name);
    }

```

However, when a null class is being created, 
```java
    /**
     * Creates a Logger for the given class name.
     */
    public static <T> Logger getLogger(Class<T> clazz) {
        if (clazz == null) {
            System.out.println(clazz);
            return Logger.getLogger("");
        }
        return getLogger(clazz.getSimpleName());
    }
```
We can see that the default `Logger.getLogger()`  is called instead which prevents the handlers from being added.

---

We can verify this by modifying a logger.
```java
    public static final Version VERSION = new Version(0, 2, 0, true);

    private static final Logger logger = LogsCenter.getLogger(Object.class.getSuperclass()); // from MainApp.class

    protected Ui ui;
    protected Logic logic;
    protected Storage storage;
    protected Model model;
    protected Config config;

    @Override
```
image
We can see that logs of null class are not being properly logged.

---

As of now, no null loggers have been created but I think this should still be considered as a bug? or that the argument that null loggers logs should not be logged in the log file.

Eclipse-Dominator:
Fixes #173 

There is a check when a null class logger is created. 
However, AB3 currently do not actually use null class loggers.

Let's replace the null class check with assertion instead as per YAGNI principe.

### DIFF
--- a/src/main/java/seedu/address/commons/core/LogsCenter.java
+++ b/src/main/java/seedu/address/commons/core/LogsCenter.java
@@ -1,5 +1,7 @@
 package seedu.address.commons.core;
 
+import static java.util.Objects.requireNonNull;
+
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.logging.ConsoleHandler;
@@ -53,9 +55,7 @@ public class LogsCenter {
      * Creates a Logger for the given class name.
      */
     public static <T> Logger getLogger(Class<T> clazz) {
-        if (clazz == null) {
-            return Logger.getLogger("");
-        }
+        requireNonNull(clazz);
         return getLogger(clazz.getSimpleName());
     }
 


damithc (changes requested):
The commit message raises some questions in the reader's mind: 
* What is a 'null class logger'?
* What is the 'incorrect method'? Why it is incorrect?

So, although the commit itself is small, looks like it needs a more detailed commit message.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -53,9 +53,7 @@ public static Logger getLogger(String name) {
      * Creates a Logger for the given class name.
      */
     public static <T> Logger getLogger(Class<T> clazz) {
-        if (clazz == null) {
-            return Logger.getLogger("");
-        }
+        assert clazz != null;
```
damithc: I remember we use a different method (rather that `assert`) to check for null parameters elsewhere in AB3. If that is the case, we can do the same, for consistency.

damithc (changes requested):
The current commit message is more informative than before. But a few more details are still missing.

Something like the below? Tweak as you see fit.
```
Add a null check to LogCenter.getLogger(Class<T>)

LogCenter.getLogger(Class<T> clazz) creates a Logger named
after the given clazz. If the clazz is null, it is supposed to
create a Logger with an empty String as the name. 

The current implementation calls Logger.getLogger("")
instead of LogCenter.getLogger(""), which results in the Logger
object not having the handlers that are added inside the
LogCenter.getLogger(String) method. This means log messages
written using such a logger will not appear in the log files.

However, the current codebase has no use case for passing a
null clazz to this method. As per the YAGNI principle,
we should not implement that ability until we actually need it.

Hence, instead of correcting the said error, let's add an assertion
to ensure null is never passed to this method, eliminating the need
for the buggy code path in the first place. 
```

Eclipse-Dominator:
I have replaced certain instance of clazz with class in the commit since clazz is a term to replace class in java to avoid using preserved keyword.

damithc (approved):

PR #175, opened 2023-05-14
[#171] Updating gitignore to account for bin folder

### LINKED ISSUE #171: Updating gitignore to account for bin folder
Currently, running `./gradlew` (not `./gradlew build`) in the terminal will generate a bin folder in the current directory which contains all the .class files. This folder should be included in .gitignore.

Eclipse-Dominator:
Fixes #171 

Previously, running the command `./gradlew` generated a bin folder containing all the .class files.
To prevent accidentally committing files to the repository, let's adds /bin/ to the .gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Gradle build files
 /.gradle/
 /build/
+/bin/
 src/main/resources/docs/
 
 # IDEA files


damithc:
I could not reproduce this behavior on Windows i.e., I ran `gradlew` but it did not create a `bin` folder. 🤔 

Commit message: `generated` -> `generates`?

Eclipse-Dominator:
After some investigation, the creation of `bin/` folder seems to be caused by a separate vscode extension. As the focus of the module is the use of Intellij, let's close this for now.

PR #176, opened 2023-05-17
[#162] Show warning when ignoring repeated parameters

### LINKED ISSUE #162: Give a warning when a parameter is repeated in a user command

Eclipse-Dominator:
Fixes #162 

Currently, repeated occurrences of the same parameter in a command are being ignored without any notification to the user. This can lead to potential user errors going unnoticed, especially when using the wrong parameter prefix multiple times.

This commit introduces a friendly warning system that alerts the user when a parameter is being ignored. Now, if a parameter is repeated in a command, all occurrences except the last one will trigger a warning message, ensuring that the user is aware of any unintentional mistakes or typos.

Let's provide a better user experience by catching and notifying them of ignored parameters, improving the overall usability and preventing potential errors.

### DIFF
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -49,6 +49,8 @@ public class EditCommand extends Command {
     public static final String MESSAGE_EDIT_PERSON_SUCCESS = "Edited Person: %1$s";
     public static final String MESSAGE_NOT_EDITED = "At least one field to edit must be provided.";
     public static final String MESSAGE_DUPLICATE_PERSON = "This person already exists in the address book.";
+    public static final String MESSAGE_DUPLICATE_FIELD = "Detected multiple modification to a same parameter, "
+            + "applying the latest change.\n";
 
     private final Index index;
     private final EditPersonDescriptor editPersonDescriptor;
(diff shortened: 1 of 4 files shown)


src/main/java/seedu/address/logic/commands/EditCommand.java
```diff
@@ -49,6 +49,8 @@ public class EditCommand extends Command {
     public static final String MESSAGE_EDIT_PERSON_SUCCESS = "Edited Person: %1$s";
     public static final String MESSAGE_NOT_EDITED = "At least one field to edit must be provided.";
     public static final String MESSAGE_DUPLICATE_PERSON = "This person already exists in the address book.";
+    public static final String MESSAGE_DUPLICATE_FIELD = "Detected multiple modification to a same parameter, "
+            + "applying the latest change.\n";
```
damithc, line 53: modification -> modifications?
Is it correct to say _modifications_ i.e., are these strictly modifications or something else? Similarly, 'latest change' doesn't seem like the correct phrase.

damithc (changes requested):
I'm a bit hesitant about this approach because,
1. It doesn't inform the user which parameter is duplicated
2. It doesn't differentiate between one parameter being duplicated vs multiple parameters being duplicated

Perhaps we should explore alternatives to see if there is a better way?

Eclipse-Dominator:
Hmm, alright, I chose this initially as I felt this will require the least amount of changes but I will come up with a secondary approach to display which field has been duplicated.

damithc:
> Hmm, alright, I chose this initially as I felt this will require the least amount of changes but I will come up with a secondary approach to display which field has been duplicated.

Yup, this may still be the best option but it's better to explore other options before making a decision.

Eclipse-Dominator:
I have created a secondary PR to show an alternative approach in #181

Eclipse-Dominator:

damithc:
Closing in favor of #190

PR #177, opened 2023-05-18
[#166] Bump javafx to version 11.0.2

### LINKED ISSUE #166: Bump javafx to version 11.0.2
As mentioned [here](https://bugs.openjdk.org/browse/JDK-8210411). Using version 11 of javafx causes conflict with Wayland in Ubuntu 18.04 and above. Consequently, it also led to errors in students using WSLg since most WSL users by default uses a version of ubuntu. This can be fixed by bumping javafx version to 11.0.2 from 11. 

Besides bumping to 11.0.2, it is probably also better to include `/hs_err_pid[0-9]*.log` in the `.gitignore` to avoid pushing log files to the repo for potential future errors.

Eclipse-Dominator:
Fixes #166 

Conflict between JavaFx version 11 and Ubuntu 18.04 and above described in JDK-8210411 presents a compatibility issue for users using Ubuntu 18.04 and above. This fatal crash also causes JVM to generate a hs_err_pid log file.

This issue is fixed in JavaFx version 11.0.2. Let's bump JavaFx to the appropriate version to avoid this issue.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ task coverage(type: JacocoReport) {
 
 dependencies {
     String jUnitVersion = '5.4.0'
-    String javaFxVersion = '11'
+    String javaFxVersion = '11.0.2'
 
     implementation group: 'org.openjfx', name: 'javafx-base', version: javaFxVersion, classifier: 'win'
     implementation group: 'org.openjfx', name: 'javafx-base', version: javaFxVersion, classifier: 'mac'


damithc (changes requested):
Commit 1:
The commit message (subject or body) does not specify what is the actual issue being fixed. 
Include a brief description. In addition, if there is an online description of the problem, we can give a link.
`described in JDK-8210411 presents a compatibility issue for users using` is longer than 72 chars.

Commit 2:
Let's not sneak in unrelated changes into a PR. That reduces the searchability of the PR/issue tracker.

damithc (commented):
I assume you tested this to confirm the fix works (I didn't).

Separately, check our JavaFX guide (https://se-education.org/guides/tutorials/javaFx.html) and the Grade guide to see if they need updating too.

Commit message subject is too generic?

damithc:
@zhongfu I remember you did some investigation into this problem earlier. If you are free, your inputs are welcome.

zhongfu (approved):
Looks good to me; I can confirm that this PR allows AB3 to run on Wayland without needing to force the use of X11 explicitly

(p.s. sorry for the late reply -- I missed this somehow)

damithc:
@zhongfu thanks very much for checking. :+1:

damithc (approved):

PR #181, opened 2023-06-02
[#162] Show warning when ignoring repeated  (Alternative approach)

### LINKED ISSUE #162: Give a warning when a parameter is repeated in a user command

Eclipse-Dominator:
Fixes #162 
This PR provides an alternative approach which displays all the fields that are being duplicated as compared to #176 

Currently, repeated occurrences of the same parameter in a command are being ignored without any notification to the user. This can lead to potential user errors going unnoticed, especially when using the wrong parameter prefix multiple times.

This commit introduces a friendly warning system that alerts the user when a parameter is being ignored. Now, if a parameter is repeated in a command, all occurrences except the last one will trigger a warning message, ensuring that the user is aware of any unintentional mistakes or typos.

Let's provide a better user experience by catching and notifying them of ignored parameters, improving the overall usability and preventing potential errors.

Example:
Multiple fields duplicated:
image

Single field duplicated:
image

### DIFF
--- a/src/main/java/seedu/address/logic/commands/AddCommand.java
+++ b/src/main/java/seedu/address/logic/commands/AddCommand.java
@@ -1,13 +1,18 @@
 package seedu.address.logic.commands;
 
 import static java.util.Objects.requireNonNull;
+import static seedu.address.commons.core.Messages.MESSAGE_DUPLICATE_FIELDS;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_ADDRESS;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_EMAIL;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_NAME;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_PHONE;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_TAG;
 
+import java.util.Set;
+import java.util.stream.Collectors;
+
 import seedu.address.logic.commands.exceptions.CommandException;
+import seedu.address.logic.parser.Prefix;
 import seedu.address.model.Model;
 import seedu.address.model.person.Person;
 
@@ -37,13 +42,23 @@ public class AddCommand extends Command {
     public static final String MESSAGE_DUPLICATE_PERSON = "This person already exists in the address book";
 
     private final Person toAdd;
+    private final Set<String> duplicateFields;
+
+    /**
+     * Creates an AddCommand to add the specified {@code Person} with duplicated fields {@code duplicateFields}.
+     */
+    public AddCommand(Person person, Set<Prefix> duplicateFields) {
+        requireNonNull(person);
+        requireNonNull(duplicateFields);
+        toAdd = person;
+        this.duplicateFields = duplicateFields.stream().map(Prefix::toString).collect(Collectors.toSet());
+    }
 
     /**
      * Creates an AddCommand to add the specified {@code Person}
      */
     public AddCommand(Person person) {
-        requireNonNull(person);
-        toAdd = person;
+        this(person, Set.of());
     }
 
     @Override
@@ -55,13 +70,22 @@ public class AddCommand extends Command {
         }
 
         model.addPerson(toAdd);
-        return new CommandResult(String.format(MESSAGE_SUCCESS, toAdd));
+
+        String warning = "";
+
+        if (this.duplicateFields.size() > 0) {
+            warning = String.format(MESSAGE_DUPLICATE_FIELDS, duplicateFields.size() > 1 ? "s" : "",
+                    String.join(" ", this.duplicateFields));
+        }
+
+        return new CommandResult(String.format(warning + MESSAGE_SUCCESS, toAdd));
     }
 
     @Override
     public boolean equals(Object other) {
         return other == this // short circuit if same object
                 || (other instanceof AddCommand // instanceof handles nulls
-                && toAdd.equals(((AddCommand) other).toAdd));
+                && toAdd.equals(((AddCommand) other).toAdd))
+                && duplicateFields.equals(((AddCommand) other).duplicateFields);
     }
 }
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -1,6 +1,7 @@
 package seedu.address.logic.commands;
 
 import static java.util.Objects.requireNonNull;
+import static seedu.address.commons.core.Messages.MESSAGE_DUPLICATE_FIELDS;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_ADDRESS;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_EMAIL;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_NAME;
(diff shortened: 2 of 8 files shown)


src/main/java/seedu/address/logic/commands/EditCommand.java
```diff
@@ -49,6 +49,8 @@ public class EditCommand extends Command {
     public static final String MESSAGE_EDIT_PERSON_SUCCESS = "Edited Person: %1$s";
     public static final String MESSAGE_NOT_EDITED = "At least one field to edit must be provided.";
     public static final String MESSAGE_DUPLICATE_PERSON = "This person already exists in the address book.";
+    public static final String MESSAGE_DUPLICATE_FIELDS = "Detected multiple modifications for the following field%s: "
+            + "%s.\nOnly the last change will be applied.\n";
```
damithc: `multiple modifications` -> `multiple values`?

src/main/java/seedu/address/logic/parser/EditCommandParser.java
```diff
@@ -44,15 +44,27 @@ public EditCommand parse(String args) throws ParseException {
...
         if (argMultimap.getValue(PREFIX_NAME).isPresent()) {
+            if (argMultimap.getAllValues(PREFIX_NAME).size() > 1) {
+                editPersonDescriptor.addDuplicateField(PREFIX_NAME.getPrefix());
+            }
             editPersonDescriptor.setName(ParserUtil.parseName(argMultimap.getValue(PREFIX_NAME).get()));
         }
         if (argMultimap.getValue(PREFIX_PHONE).isPresent()) {
+            if (argMultimap.getAllValues(PREFIX_PHONE).size() > 1) {
+                editPersonDescriptor.addDuplicateField(PREFIX_PHONE.getPrefix());
+            }
```
damithc: This `if` block is duplicated a few times. Is there a way to avoid that duplication?

src/test/java/seedu/address/testutil/EditPersonDescriptorBuilder.java
```diff
@@ -55,6 +68,14 @@ public EditPersonDescriptorBuilder withPhone(String phone) {
         return this;
     }
 
+    /**
+     * Add a duplicate phone field to the {@code EditPersonDescriptor} that we are building.
+     */
+    public EditPersonDescriptorBuilder withDuplicatePhone() {
+        descriptor.addDuplicateField(PREFIX_PHONE.getPrefix());
+        return this;
+    }
```
damithc: Instead of adding these methods, can the same be accomplished by calling `withPhone()` multiple times?

damithc (commented):
Added some prelim comments.

src/main/java/seedu/address/logic/commands/EditCommand.java
damithc: Also, this message uses two words for the same thing `modifications`, `change`.

src/test/java/seedu/address/testutil/EditPersonDescriptorBuilder.java
Eclipse-Dominator: It's possible but it might be less intuitive to when building the test cases since calling the same method twice may not be that intuitive as compared to calling it via a specific method.

I will try to experiment combining the functionality in the same method.

damithc: I felt calling the method `withPhone()` twice is the more natural way to specify the same parameter twice. With a specific method like `withDuplicatePhone()`, you open the door for calling `withDuplicatePhone()` without calling `withPhone()` first.
Yes, experiment variations and see how it goes.

src/main/java/seedu/address/logic/parser/EditCommandParser.java
```diff
@@ -79,4 +79,14 @@ private Optional<Set<Tag>> parseTagsForEdit(Collection<String> tags) throws Pars
...
+    /**
+     * Checks if the prefix is present in the argument multimap, if the multiple prefixes are found, add them to the
+     * duplicate fields list in the descriptor.
+     */
+    private boolean ifPresentHandleDuplicate(EditPersonDescriptor descriptor, ArgumentMultimap argMap, Prefix prefix) {
+        if (argMap.getAllValues(prefix).size() > 1) {
+            descriptor.addDuplicateField(prefix.getPrefix());
+        }
+        return argMap.getValue(prefix).isPresent();
+    }
```
damithc: Good attempt at reducing duplication, but I'm not sure this method is the way to do it. It is best when a method is either a command (i.e., does something) or a query (i.e., answers a question) but not both. 
For example, someone seeing a call to this method will think it is a query only, thus misleading the reader.

Is there any way to avoid the above flaw, even at the expense of slightly more code?

src/test/java/seedu/address/testutil/EditPersonDescriptorBuilder.java
```diff
@@ -41,32 +46,40 @@ public EditPersonDescriptorBuilder(Person person) {
 
     /**
      * Sets the {@code Name} of the {@code EditPersonDescriptor} that we are building.
+     * If the name is already present, it will be added to the duplicate fields list.
      */
     public EditPersonDescriptorBuilder withName(String name) {
+        descriptor.getEmail().ifPresent(x -> descriptor.addDuplicateField(PREFIX_NAME.getPrefix()));
         descriptor.setName(new Name(name));
```
damithc: getEmail() -> getName()?
BTW, should the logic in line 52 be inside the setName() method instead?

damithc (commented):
Raised a few ideas about the implementation strategy. Might not work but worth exploring.

src/test/java/seedu/address/testutil/EditPersonDescriptorBuilder.java
Eclipse-Dominator: I missed this when copilot generated the code wrongly.

Eclipse-Dominator:
Upon some further investigation, AddCommand is also prone to having ignore parameter, as such it might be better to include the duplicated field as a separate parameter in the constructor rather than modifying the EditPersonDescriptor 

The new change shifts the warning message to Messages.
The code can further be abstracted for AddCommand and EditCommand to extend from parseCommand to reduce some level of code duplication but I feel that this might be too much of an overkill for this case.

src/main/java/seedu/address/logic/commands/AddCommand.java
```diff
@@ -37,13 +42,23 @@ public class AddCommand extends Command {
...
 
     private final Person toAdd;
+    private final Set<String> duplicateFields;
 
     /**
-     * Creates an AddCommand to add the specified {@code Person}
+     * Creates an AddCommand to add the specified {@code Person} with duplicated fields {@code duplicateFields}.
      */
-    public AddCommand(Person person) {
+    public AddCommand(Person person, Set<Prefix> duplicateFields) {
```
damithc, line 50: I don't think this is a good approach. AddCommand should not need to know about duplicate fields.

damithc (commented):
It looks like adding this warning is creating unnecessary complications. The alternative is to remove that feature altogether i.e., repeated parameters will be considered an error. Can you put this on hold for now and create a PR to remove that feature? After that, we can compare pros/cons of the two ways to add a warning and removing the feature entirely.

src/main/java/seedu/address/logic/commands/AddCommand.java
Eclipse-Dominator: Maybe we can make AddCommand throw exception for duplicate fields?
and make EditCommand show the relevant warning? 

That way we can also ground duplicate field to the EditPersonDescriptor

damithc: A non-trivial design change is required to accommodate this warning message. The resulting design will be more complicated than the current one. Perhaps it is an acceptable cost if we want to optimize the application for a better user experience but AB3 also has to optimize for novice programmers. So, I think we may have to strip out this functionality altogether, for the sake of keeping the code tenable for novice programmers.

Eclipse-Dominator:
> It looks like adding this warning is creating unnecessary complications. The alternative is to remove that feature altogether i.e., repeated parameters will be considered an error. Can you put this on hold for now and create a PR to remove that feature? After that, we can compare pros/cons of the two ways to add a warning and removing the feature entirely.

See #190

Eclipse-Dominator:

damithc:
Closing in favor of #190

PR #185, opened 2023-06-02
[#182] Local github page deployment fails on Ruby >=3.0

### LINKED ISSUE #182: Local github page deployment fails on Ruby >=3.0
When following instructions to test github page deployment locally [here](https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/testing-your-github-pages-site-locally-with-jekyll)

The following errors is seen after running `bundle exec jekyll serve`
```
gems/gems/kramdown-1.17.0/lib/kramdown/parser/html.rb:10:in `require': cannot load such file -- rexml/parsers/baseparser (LoadError)
```

This issue is due to some compatibility errors with Ruby 3.0 and Jekyll, and the general approach is to add "webrick" package to gemfile seen [here](https://github.com/jekyll/jekyll/issues/8523)

However, the gemfiles versions in the Gemfile.lock is too low and the error above will still persist. 
To ensure that the local deployment works, the packages still needs to be updated via `bundle update`. 

Currently, there are 2 approaches to resolving this issue

---
Adding "webrick" to `Gemfile`
Update `Gemfile.lock` via `bundle update`

---
Adding "webrick" to `Gemfile`
Delete and add `Gemfile.lock` to `.gitignore`

---

Adding "webrick" is a necessary step for ruby versions >=3.0 in order for local testing to work.

Eclipse-Dominator:
Fixes #182 

Running `bundle exec jekyll serve` to test GitHub page deployment will fail due to a compatibility issue with Jekyll 3 and Ruby >=3.0.

Let's resolve this by
- adding "webrick" to Gemfile
- updating the relevant packages


damithc (commented):
Looks OK to me, but I am not very familiar with Ruby or Jekyll. Let's leave it here for a few days in case others have any comments.

Commit message: 
* Try to make the commit subject more specific so that it tell us what exact problem was fixed.
* The local testing is nothing to do with GitHub Pages. It's just a Jekyll server.

damithc:
No further comments on the fix but a couple of questions, mostly out of curiosity.
* Why does adding webricks fix the problem?
* How did you decide those packages need to be upgraded to those exact versions?

Eclipse-Dominator:
I think prior to ruby 3.0, "webricks" package is installed by default. That's why after ruby 3.0, jekyll server will fail since they can no longer find the webricks package. 

I think those were the version numbers after I updated ruby packages. I think I can fine tune the versions more? Alternatively, just removing gem.lock file and let the users install the latest version will work as well

damithc:
I have no further comments on this PR. 
I also don't know enough about Jekyll to decide whether we should keep `gem.lock` in the repo or remove it.
@se-edu/tech-team-level1 feel free to weigh in.

dlimyy:
I tested it on my end and LGTM! I think it is better to keep the `gem.lock` file just in case the newer versions of the gems installed by bundler causes the website to crash.

dlimyy (commented):
I tested it on my end and LGTM! I think it is better to keep the gem.lock file just in case the newer versions of the gems installed by bundler causes the website to crash.

damithc:
Thanks for the fix @Eclipse-Dominator 
Note that I edited the commit message slightly, to explain the reason for the crash so that the reason for the change is evident.

PR #186, opened 2023-06-02
[#167] Unnecessary catch block in MainApp and Exceptions in Interfaces

### LINKED ISSUE #167: Unnecessary catch block in MainApp and Exceptions in Interfaces

Eclipse-Dominator:
Fixes #167 

Alternative Approach: #191 

Classes that implements the methods,
- readAddressBook() from AddressBookStorage interface
- readUserPrefs() from UserPrefsStorage interface does not actually throw any IOException as such exceptions are caught within the method and is being treated as DataConversionException which may not be the case.

All IOException are caught by JsonUtil.readJsonFile() which is called by classes implementing these methods.

This method also causes ConfigUtil.readConfig() to not throw IOExceptions when handling actual IOExceptions that is not caused by a conversion error.

This causes the wrong log messages to be logged when an actual IOException occur.

Let's
- decouple actual conversion errors from IOExceptions and restore functionality to the catch blocks in MainApp.java that handles IOExceptions.
- add catch blocks to handle IOException when reading config files

---

We can see the effect of the change below when reading files that does not have sufficient read permissions:

image

---

Right now, actual IOExceptions (not conversion errors) are still being handled like conversion errors where a blank/default save file is used instead. 
I would prefer to show some form of warning then exit the application instead. However, this might require additional changes to UI behavior so I left it as such for now.

### DIFF
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -84,11 +84,12 @@ public class MainApp extends Application {
             }
             initialData = addressBookOptional.orElseGet(SampleDataUtil::getSampleAddressBook);
         } catch (DataConversionException e) {
-            logger.warning("Data file at " + storage.getAddressBookFilePath() + " is not in the correct format."
+            logger.warning("Data file at " + storage.getAddressBookFilePath()
+                    + " is might be corrupted or in the wrong format."
                     + " Will be starting with an empty AddressBook.");
             initialData = new AddressBook();
         } catch (IOException e) {
-            logger.warning("Problem while reading from the file " + storage.getAddressBookFilePath()
+            logger.warning("Problem while reading from the data file " + storage.getAddressBookFilePath()
                     + ". Will be starting with an empty AddressBook.");
             initialData = new AddressBook();
         }
@@ -106,10 +107,7 @@ public class MainApp extends Application {
      * if {@code configFilePath} is null.
      */
     protected Config initConfig(Path configFilePath) {
-        Config initializedConfig;
-        Path configFilePathUsed;
-
-        configFilePathUsed = Config.DEFAULT_CONFIG_FILE;
+        Path configFilePathUsed = Config.DEFAULT_CONFIG_FILE;
 
         if (configFilePath != null) {
             logger.info("Custom Config file specified " + configFilePath);
@@ -117,18 +115,22 @@ public class MainApp extends Application {
         }
 
         logger.info("Using config file : " + configFilePathUsed);
-
+        Optional<Config> configOptional = Optional.empty();
         try {
-            Optional<Config> configOptional = ConfigUtil.readConfig(configFilePathUsed);
+            configOptional = ConfigUtil.readConfig(configFilePathUsed);
             if (!configOptional.isPresent()) {
                 logger.info("Creating new config file " + configFilePathUsed);
             }
-            initializedConfig = configOptional.orElse(new Config());
         } catch (DataConversionException e) {
-            logger.warning("Config file at " + configFilePathUsed + " is not in the correct format."
-                    + " Using default config properties.");
-            initializedConfig = new Config();
+            logger.warning("Config file at " + configFilePathUsed
+                    + " might be corrupted or in the wrong format"
+                    + ". Using default config properties.");
+        } catch (IOException e) {
+            logger.warning("Problem while reading from the config file " + configFilePathUsed
+                    + ". Using default config properties.");
         }
+
+        Config initializedConfig = configOptional.orElse(new Config());
 
         //Update config file in case it was missing to begin with or there are new/unused fields
         try {
@@ -148,22 +150,20 @@ public class MainApp extends Application {
         Path prefsFilePath = storage.getUserPrefsFilePath();
         logger.info("Using preference file : " + prefsFilePath);
 
-        UserPrefs initializedPrefs;
+        Optional<UserPrefs> prefsOptional = Optional.empty();
         try {
-            Optional<UserPrefs> prefsOptional = storage.readUserPrefs();
+            prefsOptional = storage.readUserPrefs();
             if (!prefsOptional.isPresent()) {
                 logger.info("Creating new preference file " + prefsFilePath);
             }
-            initializedPrefs = prefsOptional.orElse(new UserPrefs());
         } catch (DataConversionException e) {
-            logger.warning("Preference file at " + prefsFilePath + " is not in the correct format."
+            logger.warning("Preference file at " + prefsFilePath + " might be corrupted or in the wrong format."
                     + " Using default preferences.");
-            initializedPrefs = new UserPrefs();
         } catch (IOException e) {
             logger.warning("Problem while reading from preference file " + prefsFilePath
-                    + ". Will be starting with an empty AddressBook.");
-            initializedPrefs = new UserPrefs();
+                    + ". Using default preferences");
         }
+        UserPrefs initializedPrefs = prefsOptional.orElse(new UserPrefs());
 
         //Update prefs file in case it was missing to begin with or there are new/unused fields
         try {
--- a/src/main/java/seedu/address/commons/util/JsonUtil.java
+++ b/src/main/java/seedu/address/commons/util/JsonUtil.java
@@ -55,8 +55,8 @@ public class JsonUtil {
      * @param classOfObjectToDeserialize Json file has to correspond to the structure in the class given here.
      * @throws DataConversionException if the file format is not as expected.
      */
-    public static <T> Optional<T> readJsonFile(
-            Path filePath, Class<T> classOfObjectToDeserialize) throws DataConversionException {
+    public static <T> Optional<T> readJsonFile(Path filePath, Class<T> classOfObjectToDeserialize)
+            throws DataConversionException, IOException {
         requireNonNull(filePath);
 
         if (!Files.exists(filePath)) {
@@ -68,9 +68,12 @@ public class JsonUtil {
 
         try {
             jsonFile = deserializeObjectFromJsonFile(filePath, classOfObjectToDeserialize);
+        } catch (JsonProcessingException e) {
+            logger.warning("Error processing data from jsonFile file " + filePath + ": " + e);
+            throw new DataConversionException(e);
         } catch (IOException e) {
             logger.warning("Error reading from jsonFile file " + filePath + ": " + e);
-            throw new DataConversionException(e);
+            throw e;
         }
 
         return Optional.of(jsonFile);
@@ -96,7 +99,8 @@ public class JsonUtil {
      * @param <T> The generic type to create an instance of
      * @return The instance of T with the specified values in the JSON string
      */
-    public static <T> T fromJsonString(String json, Class<T> instanceClass) throws IOException {
+    public static <T> T fromJsonString(String json, Class<T> instanceClass)
+            throws IOException, JsonProcessingException {
         return objectMapper.readValue(json, instanceClass);
     }
 
--- a/src/main/java/seedu/address/commons/util/ConfigUtil.java
+++ b/src/main/java/seedu/address/commons/util/ConfigUtil.java
@@ -12,7 +12,7 @@ import seedu.address.commons.exceptions.DataConversionException;
  */
 public class ConfigUtil {
 
-    public static Optional<Config> readConfig(Path configFilePath) throws DataConversionException {
+    public static Optional<Config> readConfig(Path configFilePath) throws DataConversionException, IOException {
         return JsonUtil.readJsonFile(configFilePath, Config.class);
     }
 
--- a/src/main/java/seedu/address/storage/JsonAddressBookStorage.java
+++ b/src/main/java/seedu/address/storage/JsonAddressBookStorage.java
@@ -32,7 +32,7 @@ public class JsonAddressBookStorage implements AddressBookStorage {
     }
 
     @Override
-    public Optional<ReadOnlyAddressBook> readAddressBook() throws DataConversionException {
+    public Optional<ReadOnlyAddressBook> readAddressBook() throws DataConversionException, IOException {
         return readAddressBook(filePath);
     }
 
@@ -42,7 +42,7 @@ public class JsonAddressBookStorage implements AddressBookStorage {
      * @param filePath location of the data. Cannot be null.
      * @throws DataConversionException if the file is not in the correct format.
      */
-    public Optional<ReadOnlyAddressBook> readAddressBook(Path filePath) throws DataConversionException {
+    public Optional<ReadOnlyAddressBook> readAddressBook(Path filePath) throws DataConversionException, IOException {
         requireNonNull(filePath);
 
         Optional<JsonSerializableAddressBook> jsonAddressBook = JsonUtil.readJsonFile(
--- a/src/main/java/seedu/address/storage/JsonUserPrefsStorage.java
+++ b/src/main/java/seedu/address/storage/JsonUserPrefsStorage.java
@@ -26,7 +26,7 @@ public class JsonUserPrefsStorage implements UserPrefsStorage {
     }
 
     @Override
-    public Optional<UserPrefs> readUserPrefs() throws DataConversionException {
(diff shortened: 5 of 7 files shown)


src/main/java/seedu/address/MainApp.java
```diff
@@ -128,6 +128,10 @@ protected Config initConfig(Path configFilePath) {
             logger.warning("Config file at " + configFilePathUsed + " is not in the correct format."
                     + " Using default config properties.");
             initializedConfig = new Config();
+        } catch (IOException e) {
+            logger.warning("Problem while reading from the file " + configFilePathUsed
+                    + ". Will be starting with default configuration");
```
damithc: Why is this part different from the `DataConversionException` catch clause?

src/main/java/seedu/address/MainApp.java
```diff
@@ -128,6 +128,10 @@ protected Config initConfig(Path configFilePath) {
...
+            initializedConfig = new Config();
```
damithc: Is it possible to combine these two catch clauses, as the action taken is the same?  

```java
        } catch (DataConversionException | IOException e) {
            logger.warning(e.getMessage + " Using default config properties.");
            initializedConfig = new Config();
        }
```

damithc (changes requested):
Some initial comments ...

src/main/java/seedu/address/MainApp.java
Eclipse-Dominator: `DataConversionException` is a custom exception that actually takes in a file path. So when `e.getMessage()` is called on it, it returns a file path instead of the exception as compared to `IOException`.

src/main/java/seedu/address/MainApp.java
Eclipse-Dominator: I transferred from the other file checks since Config initially did not have the catch block for IOException. I will standardize the phrasing.

src/main/java/seedu/address/MainApp.java
damithc: > `DataConversionException` is a custom exception that actually takes in a file path.

I don't think so. Check the code.

Eclipse-Dominator: I got confused with some other repo. 

I have checked the `getMessage` when a DataConversionException is created, it seems that such exceptions have a really long message:

image

The warning is actually already logged prior to this by `readJsonFile` 
image

so it might better if we just provide a short description of what happened instead?

Alternatively, we can specifically override the `getMessage()` for `DataConversionException` to give a description of the error since the details are already printed prior to this?

---

Or we can just change all the catch block to say error reading from file since the reason is already logged by `readJsonFile`

damithc: >so it might better if we just provide a short description of what happened instead?

How would the code look like for that option?

Eclipse-Dominator: DataConversionException: "XXX file at <file location> is might be corrupted/in the wrong format\n"
IOException: "An error occurred while reading from XXX file at <file location>: <e.getMessage()>\n"

We can perhaps combine them into a singular catch block like so:

```java
        } catch (DataConversionException | IOException e) {
            String description = "";
            if (e instanceof DataConversionException) {
                description = "Config file at " + configFilePathUsed
                        + " is might be corrupted or in the wrong format";
            } else {
                description = "Problem while reading from the file " + configFilePathUsed;
            }
            logger.warning(description + ". Using default config properties.");
            initializedConfig = new Config();
        }
```

The result will look like so:
image

damithc: `if (e instanceof DataConversionException)` defeats the purpose of combining the two `catch` clauses. In any case, I just noticed that the `initModelManager` method uses two `catch` clauses as well. So, we can do the same here.

damithc (commented):
@Eclipse-Dominator you can proceed to the next iteration of this PR. 👍

Eclipse-Dominator:
I have included some change to reduce code duplication in handling of config and pref files.

src/main/java/seedu/address/commons/util/JsonUtil.java
```diff
@@ -68,9 +68,12 @@ public static <T> Optional<T> readJsonFile(
 
         try {
             jsonFile = deserializeObjectFromJsonFile(filePath, classOfObjectToDeserialize);
+        } catch (JsonProcessingException e) {
+            logger.warning("Error processing data from jsonFile file " + filePath + ": " + e);
```
damithc, line 72:
```suggestion
            logger.warning("Error processing data from JSON file " + filePath + ": " + e);
```

damithc (commented):
Pointed a minor nit.

Good job writing a detailed commit message. I think the key statement is the following.
>This causes the wrong log messages to be logged when an actual
>IOException occur.

We should lead with that, and elaborate more. What kind of wrong message is logged in what kind of situuation? So, the commit message can start like this:

> When X happens, the incorrect message Y is logged.
>
> Briefly describe why the above happens.
>
> Briefly describe how this PR fixes it.

damithc:
@se-edu/tech-team-level1 for your reviews please ...

src/main/java/seedu/address/commons/util/JsonUtil.java
```diff
@@ -68,9 +68,12 @@ public static <T> Optional<T> readJsonFile(
...
+            throw new DataConversionException(e);
         } catch (IOException e) {
             logger.warning("Error reading from jsonFile file " + filePath + ": " + e);
```
mayuanxin1234, line 75:
```suggestion
            logger.warning("Error reading from JSON file " + filePath + ": " + e);
```

src/main/java/seedu/address/commons/util/JsonUtil.java
```diff
@@ -68,9 +68,12 @@ public static <T> Optional<T> readJsonFile(
...
-            throw new DataConversionException(e);
+            throw e;
```
SPWwj, line 76: I've noticed a potential issue with the error handling strategy in this code. Specifically, the practice of catching and immediately re-throwing the same error could lead to some inefficiencies.

In your current setup, you've employed a logger to document exceptions. However, by re-throwing the same exception, it's possible that another component, such as MainApp, might process this exception once more. This could inadvertently lead to duplicate entries in our logs, complicating the debugging process and adding unnecessary clutter to our log files.

To enhance this, I'd suggest revisiting the error handling strategy.

SPWwj (commented):
Upon reviewing this Pull Request (PR), it appears that it may not adequately address the problem outlined in issue #167. Specifically, the changes seem to make the IOException catch block mandatory in MainApp, which may not be the intended solution.

I would recommend considering a different approach to resolve this issue. If this PR is addressing a different problem, it might be more appropriate to create a new issue to track and discuss these changes separately. This will help maintain clarity and focus in our issue tracking.


image
I feel the issue at hand is more related to the appropriateness of the error message rather than incorrect error handling.

src/main/java/seedu/address/MainApp.java
```diff
@@ -84,11 +84,12 @@ private Model initModelManager(Storage storage, ReadOnlyUserPrefs userPrefs) {
             }
             initialData = addressBookOptional.orElseGet(SampleDataUtil::getSampleAddressBook);
         } catch (DataConversionException e) {
-            logger.warning("Data file at " + storage.getAddressBookFilePath() + " is not in the correct format."
+            logger.warning("Data file at " + storage.getAddressBookFilePath()
+                    + " is might be corrupted or in the wrong format."
```
chia-yh, line 88:
```suggestion
                    + " might be corrupted or in the wrong format."
```

src/main/java/seedu/address/MainApp.java
```diff
@@ -106,30 +107,31 @@ private void initLogging(Config config) {
...
-            initializedConfig = configOptional.orElse(new Config());
         } catch (DataConversionException e) {
-            logger.warning("Config file at " + configFilePathUsed + " is not in the correct format."
-                    + " Using default config properties.");
-            initializedConfig = new Config();
+            logger.warning("Config file at " + configFilePathUsed
+                    + " might be corrupted or in the wrong format"
+                    + ". Using default config properties.");
+        } catch (IOException e) {
+            logger.warning("Problem while reading from the config file " + configFilePathUsed
```
jundatan, line 129:
```suggestion
            logger.warning("Problem while reading from the config file at " + configFilePathUsed
```

jundatan (commented):
Suggested change for warning logger.

src/main/java/seedu/address/commons/util/JsonUtil.java
damithc: Good observations @SPWwj 
I wouldn't worry about the efficiency so much though. The difference in performance is unlikely to matter.
There is value is logging an exception along multiple points between where it was thrown and where it was handled. For one thing, a similar exception can be thrown from multiple places but handled in one place, which means it is useful to record where it was thrown. For another, the decision to log or not is best made locally (i.e., without considering where the method is being used) -- as the usage of a method should be able to change freely without having to change the method. So, we shouldn't make any assumptions about how the method will be used and how the exception will be handled over there.

damithc:
>Upon reviewing this Pull Request (PR), it appears that it may not adequately address the problem outlined in issue https://github.com/se-edu/addressbook-level3/issues/167. Specifically, the changes seem to make the IOException catch block mandatory in MainApp, which may not be the intended solution.

I agree what what @SPWwj  said above in that there is disparity between the issue and the PR. The original intention was to remove the unnecessary catch block.  I prefer that approach as it simplifies the code without any loss of value (because we do not currently have different handlings for incorrect format in JSON files vs other I/O problems with the JSON files). 

In addition, The name `DataConversionException` may be misleading because we use it to represent both kind of problems mentioned above. Perhaps something like `ConfigLoadingException` is more representative?

SPWwj:
> `DataConversionException`

The renaming of the exception to ConfigLoadingException aligns well with its purpose.

In JsonUtil.java, handling the IOException by logging it provides valuable information about the location of the error, aiding in debugging.

In MainApp, handling the ConfigLoadingException manages any issues that arise during the data fetching process, enhancing the robustness of error handling and preventing application crashes due to unhandled exceptions.

The use of interfaces for the readUserPrefs method allows for various implementations of the logic. It's not restricted to reading data from a local JSON file; it could also fetch data over the internet. As a result, different types of exceptions, not just IOException, could be encountered. By using ConfigLoadingException as a general exception, these scenarios can be handled in a unified way.

src/main/java/seedu/address/MainApp.java
```diff
@@ -148,22 +150,20 @@ protected UserPrefs initPrefs(UserPrefsStorage storage) {
         Path prefsFilePath = storage.getUserPrefsFilePath();
         logger.info("Using preference file : " + prefsFilePath);
```
linustws, line 151:
```suggestion
        logger.info("Using preference file at " + prefsFilePath);
```

Eclipse-Dominator:
> > Upon reviewing this Pull Request (PR), it appears that it may not adequately address the problem outlined in issue #167. Specifically, the changes seem to make the IOException catch block mandatory in MainApp, which may not be the intended solution.
> 
> I agree what what @SPWwj said above in that there is disparity between the issue and the PR. The original intention was to remove the unnecessary catch block. I prefer that approach as it simplifies the code without any loss of value (because we do not currently have different handlings for incorrect format in JSON files vs other I/O problems with the JSON files).
> 
> In addition, The name `DataConversionException` may be misleading because we use it to represent both kind of problems mentioned above. Perhaps something like `ConfigLoadingException` is more representative?

Hmm, it's actually possible to remove the unnecessary catch block as indicated in the original PR as technically IO related errors/logs can be handled within the `jsonutil` method and we only need to mention that something went wrong while trying to read XXX file in the `MainApp`.

My original concern is that students may want to implement additional features/error handling when encountering these errors. Thus, I went ahead with restoring the functionality of the IOException catch block. 

As this is also another possibility, I will create a separate PR removing the catch blocks as intended in the original issue to explore other approaches to this. You can find the alternative approach in #191

damithc:
Closing this in favor of #191

PR #187, opened 2023-06-02
[#126] Changing the log level in config.json does not work

### LINKED ISSUE #126: Changing the log level in config.json does not work

Eclipse-Dominator:
Fixes #126 

The current way of initializing loggers causes all loggers to default to `INFO` level, even if a different level is specified via the `config.json` or programmatically. 

Let's create a logger named `ab3` to serve as the base logger, and automatically prefix all loggers created using `LogCenter#getLogger(...)` with `ab3.` so that they all become descendants of the base logger. This allows us to change the logging level of all loggers simply by changing
the logging level of the base logger (unless those loggers have specified their own logging level).

Implications: Loggers not created using the `LogCenter#getLogger(...)` methods will not come under the above control mechanism unless their name starts with `ab3.`.

---
A simple demo for `log level = FINEST` 
image

A simple demo for `log level = WARNING` 
image

---

There is a small change in the behavior of init, previously, changing the log level will cause all *subsequent* loggers to have their log level changed, however, currently, this implementation changes all loggers created from the application. 

However, some classes like AddressBook etc are created before log level is being updated and I am afraid that this might defeat the purpose of changing the log levels. 

---
Another minor thing I noticed: [comment](https://github.com/se-edu/addressbook-level3/pull/187#issuecomment-1579559394)

---

The current code base mostly uses `INFO` and I believe it might be more helpful if we reclassify some of the logs to use `warning` and higher like changing logs caused by incorrect command to `WARN` and logs caused by IOException to `FATAL/ERROR`.

### DIFF
--- a/src/main/java/seedu/address/commons/core/LogsCenter.java
+++ b/src/main/java/seedu/address/commons/core/LogsCenter.java
@@ -21,34 +21,41 @@ public class LogsCenter {
     private static final int MAX_FILE_COUNT = 5;
     private static final int MAX_FILE_SIZE_IN_BYTES = (int) (Math.pow(2, 20) * 5); // 5MB
     private static final String LOG_FILE = "addressbook.log";
+    private static final Logger logger; // logger for this class
+    private static Logger baseLogger; // to be used as the parent of all other loggers created by this class.
     private static Level currentLogLevel = Level.INFO;
-    private static final Logger logger = LogsCenter.getLogger(LogsCenter.class);
-    private static FileHandler fileHandler;
-    private static ConsoleHandler consoleHandler;
 
-    /**
-     * Initializes with a custom log level (specified in the {@code config} object)
-     * Loggers obtained *AFTER* this initialization will have their logging level changed<br>
-     * Logging levels for existing loggers will only be updated if the logger with the same name
-     * is requested again from the LogsCenter.
-     */
-    public static void init(Config config) {
-        currentLogLevel = config.getLogLevel();
-        logger.info("currentLogLevel: " + currentLogLevel);
+    // This static block ensures essential loggers are created early
+    static {
+        setBaseLogger();
+        logger = LogsCenter.getLogger(LogsCenter.class);
     }
 
     /**
-     * Creates a logger with the given name.
+     * Initializes loggers with the log level specified in the {@code config} object. Applies to all loggers created
+     * using {@link #getLogger(String)} and {@link #getLogger(Class)} methods except for those that are manually set.
+     */
+    public static void init(Config config) {
+        currentLogLevel = config.getLogLevel();
+        logger.info("Log level will be set as: " + currentLogLevel);
+        // set the level of the baseLogger which will be inherited by other loggers
+        baseLogger.setLevel(currentLogLevel);
+    }
+
+    /**
+     * Creates a logger with the given name prefixed by the {@code baseLogger}'s name so that the created logger
+     * becomes a descendant of the {@code baseLogger}. Furthermore, the returned logger will have the same log handlers
+     * as the {@code baseLogger}.
      */
     public static Logger getLogger(String name) {
-        Logger logger = Logger.getLogger(name);
-        logger.setUseParentHandlers(false);
-
+        // Java organizes loggers into a hierarchy based on their names (using '.' as a separator, similar to how Java
+        // packages form a hierarchy). Furthermore, loggers without a level inherit the level of their parent logger.
+        // By prefixing names of all loggers with baseLogger's name + ".", we make the baseLogger the parent of all
+        // loggers. This allows us to change the level of all loggers simply by changing the baseLogger level.
+        Logger logger = Logger.getLogger(baseLogger.getName() + "." + name);
         removeHandlers(logger);
-        addConsoleHandler(logger);
-        addFileHandler(logger);
-
-        return Logger.getLogger(name);
+        logger.setUseParentHandlers(true);
+        return logger;
     }
 
     /**
@@ -60,18 +67,7 @@ public class LogsCenter {
     }
 
     /**
-     * Adds the {@code consoleHandler} to the {@code logger}. <br>
-     * Creates the {@code consoleHandler} if it is null.
-     */
-    private static void addConsoleHandler(Logger logger) {
-        if (consoleHandler == null) {
-            consoleHandler = createConsoleHandler();
-        }
-        logger.addHandler(consoleHandler);
-    }
-
-    /**
-     * Remove all the handlers from {@code logger}.
+     * Removes all handlers from the {@code logger}.
      */
     private static void removeHandlers(Logger logger) {
         Arrays.stream(logger.getHandlers())
@@ -79,34 +75,32 @@ public class LogsCenter {
     }
 
     /**
-     * Adds the {@code fileHandler} to the {@code logger}. <br>
-     * Creates {@code fileHandler} if it is null.
+     * Creates a logger named 'ab3', containing a {@code ConsoleHandler} and a {@code FileHandler}.
+     * Sets it as the {@code baseLogger}, to be used as the parent logger of all other loggers.
      */
-    private static void addFileHandler(Logger logger) {
+    private static void setBaseLogger() {
+        baseLogger = Logger.getLogger("ab3");
+        baseLogger.setUseParentHandlers(false);
+        removeHandlers(baseLogger);
+
+        // Level.ALL is used as the level for the handlers because the baseLogger filters the log messages by level
+        // already; there is no need to control log message level of the handlers.
+
+        // add a ConsoleHandler to log to the console
+        ConsoleHandler consoleHandler = new ConsoleHandler();
+        consoleHandler.setLevel(Level.ALL);
+        baseLogger.addHandler(consoleHandler);
+
+        // add a FileHandler to log to a file
         try {
-            if (fileHandler == null) {
-                fileHandler = createFileHandler();
-            }
-            logger.addHandler(fileHandler);
+            FileHandler fileHandler = new FileHandler(LOG_FILE, MAX_FILE_SIZE_IN_BYTES, MAX_FILE_COUNT, true);
+            fileHandler.setFormatter(new SimpleFormatter());
+            fileHandler.setLevel(Level.ALL);
+            baseLogger.addHandler(fileHandler);
         } catch (IOException e) {
             logger.warning("Error adding file handler for logger.");
         }
     }
 
-    /**
-     * Creates a {@code FileHandler} for the log file.
-     * @throws IOException if there are problems opening the file.
-     */
-    private static FileHandler createFileHandler() throws IOException {
-        FileHandler fileHandler = new FileHandler(LOG_FILE, MAX_FILE_SIZE_IN_BYTES, MAX_FILE_COUNT, true);
-        fileHandler.setFormatter(new SimpleFormatter());
-        fileHandler.setLevel(currentLogLevel);
-        return fileHandler;
-    }
 
-    private static ConsoleHandler createConsoleHandler() {
-        ConsoleHandler consoleHandler = new ConsoleHandler();
-        consoleHandler.setLevel(currentLogLevel);
-        return consoleHandler;
-    }
 }


src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -20,21 +20,31 @@
...
+    private static final Logger baseLogger = Logger.getLogger(LOGGER_PREFIX);
 
     /**
-     * Initializes with a custom log level (specified in the {@code config} object)
-     * Loggers obtained *AFTER* this initialization will have their logging level changed<br>
-     * Logging levels for existing loggers will only be updated if the logger with the same name
-     * is requested again from the LogsCenter.
+     * Initializes with a custom log level (specified in the {@code config} object). <br>
+     * All log handlers will be adjust to the new level. <br>
+     * Meaning subsequent will be at the new level. <br>
```
damithc: Rephrase.
Also, use `<br>` only if strictly necessary.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -20,21 +20,31 @@
...
+     * This method should only be called once, during the start of the application.
      */
     public static void init(Config config) {
         currentLogLevel = config.getLogLevel();
         logger.info("currentLogLevel: " + currentLogLevel);
+        baseLogger.setLevel(currentLogLevel);
+        // if these handler is null, then all subsequent logger created will have the correct level
```
damithc: grammar problems

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -56,7 +66,7 @@ public static Logger getLogger(String name) {
      */
     public static <T> Logger getLogger(Class<T> clazz) {
         requireNonNull(clazz);
-        return getLogger(clazz.getSimpleName());
+        return getLogger(LOGGER_PREFIX + clazz.getSimpleName());
```
damithc: What's the need to add a prefix?

src/main/java/seedu/address/commons/core/LogsCenter.java
damithc: Also, I don't get what the comment is saying

damithc (changes requested):

src/main/java/seedu/address/commons/core/LogsCenter.java
Eclipse-Dominator: The way loggers are structured is they will all inherit from a parent logger (somewhat like a tree). 

Each logger can be represented by a path from root by separating each name with a "." character.

So for instance, logger with name "parent.name1" is a logger whose parent is a logger with name "parent".

The root logger is a logger with name "".

While it is possible to change the root logger's log level to the current setting, it might lead to slow down of the application as all package's log level might change and we only want to set the log level of loggers from AB3. 

Therefore, a prefix is added to create a parent logger for AB3 where all loggers created from `LogsCenter` will inherit from it. 

So that we can change the log levels of all AB3's loggers by changing that parent logger.

damithc: I see. Let's add that explanation above this line so that future developers will know the reason.

Eclipse-Dominator:
A consideration is whether we should have a method that creates loggers that ignore the current log level such as the logger that prints
`===[ Initializing AddressBook ]===` and `=== [ Stopping Address Book ] ===`,

damithc:
> A consideration is whether we should have a method that creates loggers that ignore the current log level such as the logger that prints
> `===[ Initializing AddressBook ]===` and `=== [ Stopping Address Book ] ===`,

As per the current PR, will they be skipped if the logging level is set to higher than `INFO`?

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -56,7 +66,16 @@ public static Logger getLogger(String name) {
...
-        return getLogger(clazz.getSimpleName());
+        /*
+         * Loggers will all inherit from a parent logger and are structured like a tree. Each logger
+         * can be represented by a path from root by separating each name with a "." character. The
+         * root logger is the parent of all loggers and is named "". Changing the log level of a
+         * logger or adding handlers to it will affect all its children loggers.
+         *
+         * As we only want to change loggers created by us, we will use a base logger named to
+         * represent the parent logger of all loggers created by us.
+         */
```
damithc: hmm... still doesn't quite explain the rationale fully.  In particular, I don't see the connection between these two:

>root logger is the parent of all loggers and is named ""

>As we only want to change loggers created by us, we will use a base logger named to  represent the parent logger of all loggers created by us.

If the objective is to ensure all 'our' loggers have a common prefix in their name, don't we already have that in the form of `seedu.address.`?

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -20,21 +20,31 @@
...
+        baseLogger.setLevel(currentLogLevel);
+
+        // Set the log level of log handlers to the current log level if they are not null
+        if (fileHandler != null) {
+            fileHandler.setLevel(currentLogLevel);
+        }
+        if (consoleHandler != null) {
+            consoleHandler.setLevel(currentLogLevel);
+        }
     }
```
damithc, line 43:
```java
    /**
     * Creates a logger with the given name.
     */
    public static Logger getLogger(String name) {
        Logger logger = Logger.getLogger(name);
        logger.setUseParentHandlers(false);
        removeHandlers(logger);
        addConsoleHandler(logger);
        addFileHandler(logger);
        return Logger.getLogger(name);
    }
```

Not related to this PR: In the code above (from this class), why do we use `return Logger.getLogger(name);` instead of `return logger;`?

damithc (commented):
Just a few questions ... to help me understand the code better.

src/main/java/seedu/address/commons/core/LogsCenter.java
Eclipse-Dominator: Logger itself statically maintain an mapping of all Loggers created. So returning logger and Logger.getLogger(name) returns the same thing.

Eclipse-Dominator:
> > A consideration is whether we should have a method that creates loggers that ignore the current log level such as the logger that prints
> > `===[ Initializing AddressBook ]===` and `=== [ Stopping Address Book ] ===`,
> 
> As per the current PR, will they be skipped if the logging level is set to higher than `INFO`?

Yes, as per current PR, this will indeed be skipped

src/main/java/seedu/address/commons/core/LogsCenter.java
damithc: Yes, I get that it is the same thing. But `return logger;` is both shorter and more natural, right?

damithc (commented):
Just a follow up question

src/main/java/seedu/address/commons/core/LogsCenter.java
damithc: @Eclipse-Dominator Can you do another small PR to fix this?

Eclipse-Dominator: Is it possible to include it within this PR since the commits will be rebased into the master branch anyway?

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -20,21 +20,31 @@
 public class LogsCenter {
     private static final int MAX_FILE_COUNT = 5;
     private static final int MAX_FILE_SIZE_IN_BYTES = (int) (Math.pow(2, 20) * 5); // 5MB
+    private static final String BASE_LOGGER_NAME = "seedu.address";
```
damithc: Let's revert back to `AB3`, to avoid confusing with package names.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -20,21 +20,31 @@
...
+     * Initializes with a custom log level (specified in the {@code config} object). All Loggers
+     * created using {@link #getLogger(String)} and {@link #getLogger(Class)} will have the same log
+     * level as the one specified in the {@code config} object.
      */
     public static void init(Config config) {
         currentLogLevel = config.getLogLevel();
         logger.info("currentLogLevel: " + currentLogLevel);
+        baseLogger.setLevel(currentLogLevel);
+
+        // Set the log level of log handlers to the current log level if they are not null
```
damithc:
```suggestion
        // set the log level of log handlers as well
```
No need to state the obvious

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -20,21 +20,31 @@
...
-     * Logging levels for existing loggers will only be updated if the logger with the same name
-     * is requested again from the LogsCenter.
+     * Initializes with a custom log level (specified in the {@code config} object). All Loggers
+     * created using {@link #getLogger(String)} and {@link #getLogger(Class)} will have the same log
+     * level as the one specified in the {@code config} object.
      */
     public static void init(Config config) {
         currentLogLevel = config.getLogLevel();
         logger.info("currentLogLevel: " + currentLogLevel);
+        baseLogger.setLevel(currentLogLevel);
```
damithc, line 42:
```suggestion
        
        // set the level of the baseLogger so that all other loggers will also be set to the same level
        baseLogger.setLevel(currentLogLevel);
```

src/main/java/seedu/address/commons/core/LogsCenter.java
damithc: It's not related to this PR at all. It will be confusing for reviewers.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -56,7 +66,15 @@ public static Logger getLogger(String name) {
      */
     public static <T> Logger getLogger(Class<T> clazz) {
         requireNonNull(clazz);
-        return getLogger(clazz.getSimpleName());
+        // The BASE_LOGGER_NAME acts as a common prefix for all application loggers,
+        // organizing them based on functionality or module.
+        // The dot (".") serves as a separator, reflecting the package structure and providing a
+        // hierarchical representation.
```
damithc: Not applicable here, as we are using just the class name (i.e., we are not using package names as logger names)

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -56,7 +66,15 @@ public static Logger getLogger(String name) {
...
+
+        // Changing the BASE_LOGGER_NAME affects all related loggers, enabling centralized control
+        // over their configuration and behavior. It allows for easy adjustment of multiple loggers
+        // simultaneously.
```
damithc: Something is not right. How does changing BASE_LOGGER_NAME (which is just a String) affects other loggers?

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -56,7 +66,15 @@ public static Logger getLogger(String name) {
...
+        return getLogger(BASE_LOGGER_NAME + "." + clazz.getSimpleName());
```
damithc: Something like this?
```
Java organizes loggers into a hierarchy based on their names (using '.' as a separator, 
similar to how Java packages form a hierarchy). Furthermore, loggers without a level inherit the
level of their parent logger.
By prefixing names of all loggers with baseLogger's name + ".", we make the 
baseLogger the parent of all loggers. This allows us to change the level of all
loggers simply by changing the baseLogger level.
```

damithc: Separately, you used `//` for this block comment, but in another PR you used `/* ... */`. Let's use one style consistently. I think `//` is preferred.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -20,21 +20,31 @@
...
+    private static final Logger baseLogger = Logger.getLogger(BASE_LOGGER_NAME);
 
     /**
-     * Initializes with a custom log level (specified in the {@code config} object)
-     * Loggers obtained *AFTER* this initialization will have their logging level changed<br>
-     * Logging levels for existing loggers will only be updated if the logger with the same name
-     * is requested again from the LogsCenter.
+     * Initializes with a custom log level (specified in the {@code config} object). All Loggers
+     * created using {@link #getLogger(String)} and {@link #getLogger(Class)} will have the same log
+     * level as the one specified in the {@code config} object.
```
damithc: A slightly more concise version:

```suggestion
     * Initializes loggers with the log level specified in the {@code config} object. Applies to
     * all loggers created using {@link #getLogger(String)} and {@link #getLogger(Class)} methods.
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -20,21 +20,31 @@
...
         logger.info("currentLogLevel: " + currentLogLevel);
+        baseLogger.setLevel(currentLogLevel);
+
+        // Set the log level of log handlers to the current log level if they are not null
+        if (fileHandler != null) {
+            fileHandler.setLevel(currentLogLevel);
+        }
+        if (consoleHandler != null) {
+            consoleHandler.setLevel(currentLogLevel);
+        }
```
damithc: BTW, can these handlers be null? If not, use assertions instead.

damithc (changes requested):
Some cosmetic suggestions

src/main/java/seedu/address/commons/core/LogsCenter.java
Eclipse-Dominator: Seems to be a typo, I meant to refer to base logger with the name of BASE_LOGGER_NAME

Eclipse-Dominator:
After some experimentation, I think using the inheritance of base loggers will lead to a much cleaner and simpler code as compared to statically storing both file and console handlers. 

The new changes here still enables the following:
- tweaking log level of child loggers will still function as expected unlike in before this PR
- In a hypothetical scenario where the parent logger is created after the child logger, the child logger will still inherit from the parent logger as long as the child logger isn't tweaked manually.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -59,15 +60,19 @@ public static <T> Logger getLogger(Class<T> clazz) {
...
     }
 
-    /**
-     * Adds the {@code consoleHandler} to the {@code logger}. <br>
-     * Creates the {@code consoleHandler} if it is null.
-     */
-    private static void addConsoleHandler(Logger logger) {
-        if (consoleHandler == null) {
-            consoleHandler = createConsoleHandler();
+    private static Logger getBaseLogger() {
```
damithc: This is a non-trivial method. Best to add a javadoc comment.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -59,15 +60,19 @@ public static <T> Logger getLogger(Class<T> clazz) {
...
+        Logger baseLogger = Logger.getLogger(BASE_LOGGER_NAME);
+        baseLogger.setUseParentHandlers(false);
+        removeHandlers(baseLogger);
+        baseLogger.addHandler(createConsoleHandler(Level.ALL));
```
damithc: Why use `Level.ALL` here?

damithc (commented):
Added a query and a suggestion.

src/main/java/seedu/address/commons/core/LogsCenter.java
Eclipse-Dominator: This is to ensure that all the handlers will log messages regardless of their level. So we dont need to track these loggers as a global variable in the class.

Since the logger is the one generating the messages, we dont want the level of the handlers to be higher than the log level of the loggers as this will cause the messages of the loggers to be ignored by the handler. (e.g. logger set to FINEST but handler is at INFO)

Since both handlers will always all messages by generated by the logger, there is no need to purposefully set the log level of each handler. Furthermore, setting these handlers to ALL won't impact performance since they are just comparing the log levels of the messages.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,14 +61,24 @@ public static <T> Logger getLogger(Class<T> clazz) {
     }
 
     /**
-     * Adds the {@code consoleHandler} to the {@code logger}. <br>
-     * Creates the {@code consoleHandler} if it is null.
+     * Creates and returns a base logger with the default configuration with the name {@code BASE_LOGGER_NAME}.
```
damithc: Also mention the handlers (as they are referenced in the last sentence). Something like this?
```suggestion
     * Creates and returns a base logger with the default configuration with the name {@code BASE_LOGGER_NAME}
     * and has a (@code ConsoleHandler} and a {@code FileHandler}.
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,14 +61,24 @@ public static <T> Logger getLogger(Class<T> clazz) {
...
+     * inherit the handlers of the base logger.
      */
-    private static void addConsoleHandler(Logger logger) {
-        if (consoleHandler == null) {
-            consoleHandler = createConsoleHandler();
+    private static Logger getBaseLogger() {
+        Logger baseLogger = Logger.getLogger(BASE_LOGGER_NAME);
+        baseLogger.setUseParentHandlers(false);
+        removeHandlers(baseLogger);
+        baseLogger.addHandler(createConsoleHandler(Level.ALL));
```
damithc: Perhaps add a note about it. Something like this?
```suggestion
        // Level.ALL is used because loggers filter the log messages by level already,
        // there is no need to control log message level of the handlers 
        baseLogger.addHandler(createConsoleHandler(Level.ALL));
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -20,35 +20,36 @@
...
-     * Logging levels for existing loggers will only be updated if the logger with the same name
-     * is requested again from the LogsCenter.
+     * Initializes loggers with the log level specified in the {@code config} object. Applies to all loggers created
+     * using {@link #getLogger(String)} and {@link #getLogger(Class)} methods except for those that are manually
+     * tweaked.
      */
     public static void init(Config config) {
         currentLogLevel = config.getLogLevel();
         logger.info("currentLogLevel: " + currentLogLevel);
+        // set the level of the baseLogger so that all other loggers will also be set to the same level
```
damithc:
```suggestion
        // set the level of the baseLogger, which will be inherited by other loggers
```

damithc (changes requested):
Looks good. Just a few cosmetic suggestions.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -20,35 +20,36 @@
 public class LogsCenter {
```
damithc: The commit message is still a bit imprecise for my liking. Is the following a good fit?
```
The current way of initializing loggers causes all loggers to
default to INFO level, even if a different level is specified
via the config.json or programmatically.

Let's create a logger named `ab3` to serve as the base logger, and 
automatically prefix all other loggers with `ab3.` so that they all
become descendants of the base logger. This allows us to change
the logging level of all loggers simply by changing the logging
level of the base logger (unless those loggers have specified
their own logging level).

Implications: Loggers not created using the LogCenter#getLogger(...)
methods will not come under the above control mechanism, unless
their name start with `ab3.`.
 ```

File: src/main/java/seedu/address/commons/core/LogsCenter.java [hunk 20]
```diff
@@ -20,35 +20,36 @@
...
     private static final int MAX_FILE_COUNT = 5;
     private static final int MAX_FILE_SIZE_IN_BYTES = (int) (Math.pow(2, 20) * 5); // 5MB
+    private static final String BASE_LOGGER_NAME = "ab3";
```
damithc: I wonder if this can be omitted. As no one needs to know this detail except the baseLogger, it feels like we are 'over-exposing' it here.
Instead, we can specify it at the point we create the base logger, and access it from the baseLogger object in other places it is needed.
Is that possible/better?

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -78,35 +93,27 @@ private static void removeHandlers(Logger logger) {
...
+     * Creates a {@code ConsoleHandler} for the log file.
+     *
+     * @param level the logging level of the console handler
+     */
+    private static ConsoleHandler createConsoleHandler(Level level) {
         ConsoleHandler consoleHandler = new ConsoleHandler();
-        consoleHandler.setLevel(currentLogLevel);
+        consoleHandler.setLevel(level);
         return consoleHandler;
     }
```
damithc: I'm wondering if these two methods can be in-lined. Although not strictly SLAP, it might reduce the amount of code without reducing readability. What do you think?

damithc (commented):
I think the work is now closer to mark. Added a few comments to probe areas in which we might be able to push it further towards a more concise/elegant solution.

src/main/java/seedu/address/commons/core/LogsCenter.java
Eclipse-Dominator: Ill try with this approach as we can retrieve the logger name via `logger::getName` but this may bring out things like "magic strings" within the function so I think exposing the name of the base logger as a function parameter might be a better solution so that we can keep all constants at the top.

src/main/java/seedu/address/commons/core/LogsCenter.java
Eclipse-Dominator: Since these methods are only referred once and are individually not very long, I think they can be inlined since we did not include any plan to create more custom loggers.

Eclipse-Dominator:
Update:
 - Made creation of handlers inline as they are only referred once
 - Moved retrieval and storage of BASE_HANDLER_NAME to within base handler
 - Updated PR description

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -22,33 +22,33 @@ public class LogsCenter {
     private static final int MAX_FILE_SIZE_IN_BYTES = (int) (Math.pow(2, 20) * 5); // 5MB
     private static final String LOG_FILE = "addressbook.log";
     private static Level currentLogLevel = Level.INFO;
+    private static final Logger baseLogger = LogsCenter.getBaseLogger("ab3");
```
damithc: We can put `"ab3"` inside the getBaseLogger method. It is OK to bury it inside the method because it has no relevance outside of that method.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -22,33 +22,33 @@ public class LogsCenter {
...
-    private static ConsoleHandler consoleHandler;
 
     /**
-     * Initializes with a custom log level (specified in the {@code config} object)
-     * Loggers obtained *AFTER* this initialization will have their logging level changed<br>
-     * Logging levels for existing loggers will only be updated if the logger with the same name
-     * is requested again from the LogsCenter.
+     * Initializes loggers with the log level specified in the {@code config} object. Applies to all loggers created
+     * using {@link #getLogger(String)} and {@link #getLogger(Class)} methods except for those that are manually
+     * tweaked.
```
damithc:
```suggestion
     * using {@link #getLogger(String)} and {@link #getLogger(Class)} methods except for those that are manually
     * set.
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +60,43 @@ public static <T> Logger getLogger(Class<T> clazz) {
...
-    }
+        // Level.ALL is used because loggers filter the log messages by level already,
+        // there is no need to control log message level of the handlers.
 
-    /**
-     * Adds the {@code fileHandler} to the {@code logger}. <br>
-     * Creates {@code fileHandler} if it is null.
-     */
-    private static void addFileHandler(Logger logger) {
+        // add handler to log to the console
```
damithc:
```suggestion
        // add a ConsoleHandler to log to the console
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +60,43 @@ public static <T> Logger getLogger(Class<T> clazz) {
...
+        ConsoleHandler consoleHandler = new ConsoleHandler();
+        consoleHandler.setLevel(Level.ALL);
+        baseLogger.addHandler(consoleHandler);
+
+        // add handler to log to a file
```
damithc:
```suggestion
        // add a FileHandler to log to a file
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +60,43 @@ public static <T> Logger getLogger(Class<T> clazz) {
...
             logger.warning("Error adding file handler for logger.");
         }
+
+        return baseLogger;
     }
 
     /**
-     * Creates a {@code FileHandler} for the log file.
-     * @throws IOException if there are problems opening the file.
+     * Remove all the handlers from {@code logger}.
```
damithc:
```suggestion
     * Removes all handlers from the {@code logger}.
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -22,33 +22,33 @@ public class LogsCenter {
     private static final int MAX_FILE_SIZE_IN_BYTES = (int) (Math.pow(2, 20) * 5); // 5MB
```
damithc, line 22: Commit message:
`prefix all other loggers with` -> `prefix all loggers created using LogCenter#getLogger(...) with`

damithc (changes requested):
Looks better now. Just a few more cosmetic suggestions.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +59,43 @@ public static <T> Logger getLogger(Class<T> clazz) {
     }
 
     /**
-     * Adds the {@code consoleHandler} to the {@code logger}. <br>
-     * Creates the {@code consoleHandler} if it is null.
+     * Creates and returns a base logger with the default configuration with the name {@code BASE_LOGGER_NAME}
+     * and has a {@code ConsoleHandler} and a {@code FileHandler}.
```
damithc: I commented on this earlier but the comment wasn't saved. Here it is again:
```suggestion
     * Creates and returns a base logger with the default configuration and containing a {@code ConsoleHandler}
     * and a {@code FileHandler}.
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -22,33 +22,32 @@ public class LogsCenter {
...
-     * Initializes with a custom log level (specified in the {@code config} object)
-     * Loggers obtained *AFTER* this initialization will have their logging level changed<br>
-     * Logging levels for existing loggers will only be updated if the logger with the same name
-     * is requested again from the LogsCenter.
+     * Initializes loggers with the log level specified in the {@code config} object. Applies to all loggers created
+     * using {@link #getLogger(String)} and {@link #getLogger(Class)} methods except for those that are manually set.
      */
     public static void init(Config config) {
         currentLogLevel = config.getLogLevel();
         logger.info("currentLogLevel: " + currentLogLevel);
```
damithc: Is this better?
```suggestion
        logger.info("log level set as: " + currentLogLevel);
```

damithc (changes requested):
Couple of things I missed out earlier

damithc:

SPWwj (approved):
image
The pull request functions as anticipated and successfully addresses the issue. The code adheres to standards, following the prescribed formatting guidelines.

LGTM!

However, I noticed in the provided screenshot that the default logging level is set to 'info', even though all the messages pertain to configuration. It might be more appropriate to set the logging level to 'config'. This observation may lead to a new issue.
image

damithc:
> However, I noticed in the provided screenshot that the default logging level is set to 'info', even though all the messages pertain to configuration. It might be more appropriate to set the logging level to 'config'. This observation may lead to a new issue.
> image

Interesting suggestion @SPWwj We can look into this separately.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +59,43 @@ public static <T> Logger getLogger(Class<T> clazz) {
     }
 
     /**
-     * Adds the {@code consoleHandler} to the {@code logger}. <br>
-     * Creates the {@code consoleHandler} if it is null.
+     * Creates and returns a base logger with the default configuration and containing a {@code ConsoleHandler}
+     * and a {@code FileHandler}.
+     * The base logger is the parent of all other loggers. Changing the level of the base logger will change
+     * the level of all other loggers if they are untweaked. Similarly, all loggers created from this class will
+     * inherit the handlers of the base logger.
```
damithc:
```suggestion
```
This part has nothing to do with this method.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +59,43 @@ public static <T> Logger getLogger(Class<T> clazz) {
     }
 
     /**
-     * Adds the {@code consoleHandler} to the {@code logger}. <br>
-     * Creates the {@code consoleHandler} if it is null.
+     * Creates and returns a base logger with the default configuration and containing a {@code ConsoleHandler}
+     * and a {@code FileHandler}.
```
damithc:
```suggestion
     * Creates and returns a logger named `ab3`, containing a {@code ConsoleHandler}
     * and a {@code FileHandler}.
```

Reason: This is the only thing the method does. It has no control over how the created logger will be used.

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -22,33 +22,32 @@ public class LogsCenter {
...
     public static void init(Config config) {
         currentLogLevel = config.getLogLevel();
-        logger.info("currentLogLevel: " + currentLogLevel);
+        logger.info("log level set as: " + currentLogLevel);
+        // set the level of the baseLogger which will be inherited by other loggers
+        baseLogger.setLevel(currentLogLevel);
     }
 
     /**
      * Creates a logger with the given name.
```
damithc: This method does a lot more than just create a logger with the given name.

```suggestion
     * Creates a logger with the given name prefixed by the {@code baseLogger}'s name so that the created
     * logger becomes a descendant of the {@code baseLogger}. Furthermore, the returned logger will have
     * the same log handlers as the {@code baseLogger}.
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -22,33 +22,32 @@ public class LogsCenter {
     private static final int MAX_FILE_SIZE_IN_BYTES = (int) (Math.pow(2, 20) * 5); // 5MB
     private static final String LOG_FILE = "addressbook.log";
     private static Level currentLogLevel = Level.INFO;
+    private static final Logger baseLogger = LogsCenter.getBaseLogger();
```
damithc:
```suggestion
    private static Level currentLogLevel = Level.INFO;
    /** The logger that will be used as the parent of all other loggers created by this class.  */
    private static final Logger baseLogger = LogsCenter.getBaseLogger();
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +59,43 @@ public static <T> Logger getLogger(Class<T> clazz) {
...
 
-    /**
-     * Remove all the handlers from {@code logger}.
-     */
-    private static void removeHandlers(Logger logger) {
-        Arrays.stream(logger.getHandlers())
-                .forEach(logger::removeHandler);
-    }
+        // Level.ALL is used because loggers filter the log messages by level already,
+        // there is no need to control log message level of the handlers.
```
damithc:
```suggestion
        // Level.ALL is used as the level or handlers because the baseLogger filters the log messages
        // by level already; there is no need to control log message level of the handlers.
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +59,43 @@ public static <T> Logger getLogger(Class<T> clazz) {
...
+     * inherit the handlers of the base logger.
      */
-    private static void addConsoleHandler(Logger logger) {
-        if (consoleHandler == null) {
-            consoleHandler = createConsoleHandler();
-        }
-        logger.addHandler(consoleHandler);
-    }
+    private static Logger getBaseLogger() {
+        Logger baseLogger = Logger.getLogger("ab3");
```
damithc: The current role of this method doesn't feel right. It returns something that is meant to be used as the baseLogger but it has no control over how it will be used.
The following feels like a better fit for what the method does.
```suggestion
    private void setBaseLogger() {
        baseLogger = Logger.getLogger("ab3");
```
It can be called inside the init() method. What do you think?

damithc (changes requested):
I found some of the comments have the right description but they appear at the wrong place. While tweaking them, I also encountered a method that may not be the right abstraction.

src/main/java/seedu/address/commons/core/LogsCenter.java
Eclipse-Dominator: I would still prefer the original method, as `init()` is only called after the config file is being read. 
This will cause the `baseLogger` to be uninitialized before then. 

To prevent this, `LogsCenter` might need to switch to using a getter method to retrieve the `baseLogger` (initialize it if it is not defined) which is not really ideal since `baseLogger` itself is a private attribute of `LogsCenter`

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +62,39 @@ public static <T> Logger getLogger(Class<T> clazz) {
...
+
+        // add a FileHandler to log to a file
         try {
-            if (fileHandler == null) {
-                fileHandler = createFileHandler();
-            }
-            logger.addHandler(fileHandler);
+            FileHandler fileHandler = new FileHandler(LOG_FILE, MAX_FILE_SIZE_IN_BYTES, MAX_FILE_COUNT, true);
+            fileHandler.setFormatter(new SimpleFormatter());
+            fileHandler.setLevel(Level.ALL);
```
SPWwj, line 98: This could be refractor as a new function private static setFileHandler()

damithc: @SPWwj Yup, if we go for strict SLAP, this can be extracted as a separate method. But in some cases it is OK to keep two levels inside a method, to reduce unnecessary method proliferation. So, a code block preceded by a comment that explains what the code block does is as good as extracting that code block as a method, provided that code block is needed only once.

src/main/java/seedu/address/commons/core/LogsCenter.java
damithc: I don't think that is a major problem, compared to an abstraction that doesn't sound right. But if you want something to execute earlier, you can use a `static` code block. See `CommandTestUtil.java` for an example. Would that work? e.g.,
```
static {
    setBaseLogger();
}
```

src/main/java/seedu/address/commons/core/LogsCenter.java
SPWwj: Occam's Razor 😂

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -22,33 +22,40 @@ public class LogsCenter {
     private static final int MAX_FILE_SIZE_IN_BYTES = (int) (Math.pow(2, 20) * 5); // 5MB
     private static final String LOG_FILE = "addressbook.log";
     private static Level currentLogLevel = Level.INFO;
-    private static final Logger logger = LogsCenter.getLogger(LogsCenter.class);
-    private static FileHandler fileHandler;
-    private static ConsoleHandler consoleHandler;
+    // The logger that will be used as the parent of all other loggers created by this class.
```
damithc: Is this comment in the right place?

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -22,33 +22,40 @@ public class LogsCenter {
...
+    private static final Logger logger;
+    private static Logger baseLogger;
+
+    static {
+        setBaseLogger();
+        logger = LogsCenter.getLogger(LogsCenter.class);
+    }
```
damithc, line 32:
```suggestion
   // This static block ensures essential loggers are created early
    static {
        setBaseLogger();
        logger = LogsCenter.getLogger(LogsCenter.class);
    }
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +67,37 @@ public static <T> Logger getLogger(Class<T> clazz) {
     }
 
     /**
-     * Adds the {@code consoleHandler} to the {@code logger}. <br>
-     * Creates the {@code consoleHandler} if it is null.
+     * Creates and returns a logger named 'ab3', and containing a {@code ConsoleHandler} and a {@code FileHandler}.
```
damithc:
```suggestion
     * Creates a logger named 'ab3', containing a {@code ConsoleHandler} and a {@code FileHandler}.
     * Sets it as the (@code baseLogger}, to be used as the parent logger of all other loggers.
```

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -22,33 +22,40 @@ public class LogsCenter {
     private static final int MAX_FILE_SIZE_IN_BYTES = (int) (Math.pow(2, 20) * 5); // 5MB
     private static final String LOG_FILE = "addressbook.log";
     private static Level currentLogLevel = Level.INFO;
-    private static final Logger logger = LogsCenter.getLogger(LogsCenter.class);
-    private static FileHandler fileHandler;
-    private static ConsoleHandler consoleHandler;
+    // The logger that will be used as the parent of all other loggers created by this class.
+    private static final Logger logger;
+    private static Logger baseLogger;
```
damithc:
```suggestion
    private static Logger baseLogger; // to be used as the parent of all loggers
    private static final Logger logger; // logger for this class
```
Perhaps just end-of-line comments (as suggested above) are enough here?
I put baseLogger above logger because I felt it is more intuitive to learn about baseLogger first. But either order is OK.

damithc: Any reason why one is final but the other is not?

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -22,33 +22,40 @@ public class LogsCenter {
...
-     * Loggers obtained *AFTER* this initialization will have their logging level changed<br>
-     * Logging levels for existing loggers will only be updated if the logger with the same name
-     * is requested again from the LogsCenter.
+     * Initializes loggers with the log level specified in the {@code config} object. Applies to all loggers created
+     * using {@link #getLogger(String)} and {@link #getLogger(Class)} methods except for those that are manually set.
      */
     public static void init(Config config) {
         currentLogLevel = config.getLogLevel();
-        logger.info("currentLogLevel: " + currentLogLevel);
+        logger.info("log level set as: " + currentLogLevel);
```
damithc:
```suggestion
        logger.info("log level will be set as: " + currentLogLevel);
```
Because it is not set yet. Alternatively, move to the bottom of the method (i.e., _after_ setting the log level).

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -60,53 +67,37 @@ public static <T> Logger getLogger(Class<T> clazz) {
...
-        return fileHandler;
-    }
-
-    private static ConsoleHandler createConsoleHandler() {
-        ConsoleHandler consoleHandler = new ConsoleHandler();
-        consoleHandler.setLevel(currentLogLevel);
-        return consoleHandler;
+    private static void removeHandlers(Logger logger) {
+        Arrays.stream(logger.getHandlers())
+                .forEach(logger::removeHandler);
```
damithc: Minor: seems this method was simply moved. Was that necessary? If not, can move back so that the commit becomes less noisy.

damithc (commented):
Just a few cosmetic suggestions.

damithc:

@SPWwj the code has changed since your last review.

damithc (approved):
No further comments from me. Thanks for the persistence with this PR @Eclipse-Dominator

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -21,34 +21,41 @@ public class LogsCenter {
...
-     * Loggers obtained *AFTER* this initialization will have their logging level changed<br>
-     * Logging levels for existing loggers will only be updated if the logger with the same name
-     * is requested again from the LogsCenter.
+     * Initializes loggers with the log level specified in the {@code config} object. Applies to all loggers created
+     * using {@link #getLogger(String)} and {@link #getLogger(Class)} methods except for those that are manually set.
      */
     public static void init(Config config) {
         currentLogLevel = config.getLogLevel();
-        logger.info("currentLogLevel: " + currentLogLevel);
+        logger.info("log level will be set as: " + currentLogLevel);
```
SPWwj:
```suggestion
        logger.info("Log level will be set as: " + currentLogLevel);
```

SPWwj (changes requested):
LGTM, just some nitty-gritty changes to make the message more consistent.

PR #189, opened 2023-06-03
[#188] Adopt a uniform `toString()` format

### LINKED ISSUE #188: Adopt a uniform `toString()` format
Taken from [AB4#220](https://github.com/se-edu/addressbook-level4/issues/220)

> We currently have all kinds of `toString()` string representations in our code base. Making them uniform will help us easily identify object string representations and their containing data in places like logs, JUnit test failure messages etc.
> 
> Note, however, that some user-facing messages depend on the `toString()` representations. These should be migrated to use formatting methods that are specialized for user-facing messages instead.

Initial investigation
--- 
Only `Person` class's `toString()` is used in user-facing messages.

Eclipse-Dominator:
Fixes #188 

Adapts parts of solution from AB4#974(https://github.com/se-edu/addressbook-level4/pull/974)

Part of the PR that includes renames, file changes and relocation of variables unrelated to `toString()` have been left out in this PR.

As they can part of another issue

### DIFF
--- a/src/main/java/seedu/address/commons/core/Messages.java
+++ b/src/main/java/seedu/address/commons/core/Messages.java
@@ -1,5 +1,7 @@
 package seedu.address.commons.core;
 
+import seedu.address.model.person.Person;
+
 /**
  * Container for user visible messages.
  */
@@ -9,5 +11,22 @@ public class Messages {
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
     public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+
+    /**
+     * Formats the {@code person} for display to the user.
+     */
+    public static String format(Person person) {
+        final StringBuilder builder = new StringBuilder();
+        builder.append(person.getName())
+                .append("; Phone: ")
+                .append(person.getPhone())
+                .append("; Email: ")
+                .append(person.getEmail())
+                .append("; Address: ")
+                .append(person.getAddress())
+                .append("; Tags: ");
+        person.getTags().forEach(builder::append);
+        return builder.toString();
+    }
 
 }
--- a/src/main/java/seedu/address/commons/util/ToStringBuilder.java
+++ b/src/main/java/seedu/address/commons/util/ToStringBuilder.java
@@ -1,0 +1,53 @@
+package seedu.address.commons.util;
+
+/**
+ * Builds a string representation of an object that is suitable as the return value of {@link Object#toString()}.
+ */
+public class ToStringBuilder {
+    private static final String OBJECT_PREFIX = "{";
+    private static final String OBJECT_SUFFIX = "}";
+    private static final String FIELD_SEPARATOR = ", ";
+    private static final String FIELD_NAME_VALUE_SEPARATOR = "=";
+
+    private final StringBuilder stringBuilder = new StringBuilder();
+    private boolean hasField;
+
+    /**
+     * Constructs a {@code ToStringBuilder} whose formatted output will be prefixed with {@code objectName}.
+     */
+    public ToStringBuilder(String objectName) {
+        stringBuilder.append(objectName).append(OBJECT_PREFIX);
+    }
+
+    /**
+     * Constructs a {@code ToStringBuilder} whose formatted output will be prefixed with the
+     * canonical class name of {@code object}.
+     */
+    public ToStringBuilder(Object object) {
+        this(object.getClass().getCanonicalName());
+    }
+
+    /**
+     * Adds a field name/value pair to the output string.
+     *
+     * @param fieldName The name of the field.
+     * @param fieldValue The value of the field.
+     * @return A reference to this {@code ToStringBuilder} object, allowing method calls to be chained.
+     */
+    public ToStringBuilder add(String fieldName, Object fieldValue) {
+        if (hasField) {
+            stringBuilder.append(FIELD_SEPARATOR);
+        }
+        stringBuilder.append(fieldName).append(FIELD_NAME_VALUE_SEPARATOR).append(fieldValue);
+        hasField = true;
+        return this;
+    }
+
+    /**
+     * Returns the built formatted string representation.
+     */
+    @Override
+    public String toString() {
+        return stringBuilder.toString() + OBJECT_SUFFIX;
+    }
+}
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -17,6 +17,7 @@ import java.util.Set;
 import seedu.address.commons.core.Messages;
 import seedu.address.commons.core.index.Index;
 import seedu.address.commons.util.CollectionUtil;
+import seedu.address.commons.util.ToStringBuilder;
 import seedu.address.logic.commands.exceptions.CommandException;
 import seedu.address.model.Model;
 import seedu.address.model.person.Address;
@@ -120,6 +121,14 @@ public class EditCommand extends Command {
                 && editPersonDescriptor.equals(e.editPersonDescriptor);
     }
 
+    @Override
+    public String toString() {
+        return new ToStringBuilder(this)
+                .add("index", index)
+                .add("editPersonDescriptor", editPersonDescriptor)
+                .toString();
+    }
+
     /**
      * Stores the details to edit the person with. Each non-empty field value will replace the
      * corresponding field value of the person.
@@ -221,6 +230,17 @@ public class EditCommand extends Command {
                     && getEmail().equals(e.getEmail())
                     && getAddress().equals(e.getAddress())
                     && getTags().equals(e.getTags());
+        }
+
+        @Override
+        public String toString() {
+            return new ToStringBuilder(this)
+                    .add("name", name)
+                    .add("phone", phone)
+                    .add("email", email)
+                    .add("address", address)
+                    .add("tags", tags)
+                    .toString();
         }
     }
 }
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -84,7 +84,7 @@ public class EditCommand extends Command {
 
         model.setPerson(personToEdit, editedPerson);
         model.updateFilteredPersonList(PREDICATE_SHOW_ALL_PERSONS);
-        return new CommandResult(String.format(MESSAGE_EDIT_PERSON_SUCCESS, editedPerson));
+        return new CommandResult(String.format(MESSAGE_EDIT_PERSON_SUCCESS, Messages.format(editedPerson)));
     }
 
     /**
--- a/src/test/java/seedu/address/AppParametersTest.java
+++ b/src/test/java/seedu/address/AppParametersTest.java
@@ -37,6 +37,13 @@ public class AppParametersTest {
         assertEquals(expected, AppParameters.parse(parametersStub));
     }
 
+    @Test
+    public void toStringMethod() {
+        AppParameters appParameters = new AppParameters();
+        String expected = AppParameters.class.getCanonicalName() + "{configPath=" + appParameters.getConfigPath() + "}";
+        assertEquals(expected, appParameters.toString());
+    }
+
     private static class ParametersStub extends Application.Parameters {
         private Map<String, String> namedParameters = new HashMap<>();
 
--- a/src/test/java/seedu/address/logic/commands/FindCommandTest.java
+++ b/src/test/java/seedu/address/logic/commands/FindCommandTest.java
@@ -74,6 +74,14 @@ public class FindCommandTest {
         assertEquals(Arrays.asList(CARL, ELLE, FIONA), model.getFilteredPersonList());
     }
 
+    @Test
+    public void toStringMethod() {
+        NameContainsKeywordsPredicate predicate = new NameContainsKeywordsPredicate(Arrays.asList("keyword"));
+        FindCommand findCommand = new FindCommand(predicate);
+        String expected = FindCommand.class.getCanonicalName() + "{predicate=" + predicate + "}";
+        assertEquals(expected, findCommand.toString());
+    }
+
     /**
      * Parses {@code userInput} into a {@code NameContainsKeywordsPredicate}.
      */
--- a/src/test/java/seedu/address/model/AddressBookTest.java
+++ b/src/test/java/seedu/address/model/AddressBookTest.java
@@ -83,6 +83,12 @@ public class AddressBookTest {
         assertThrows(UnsupportedOperationException.class, () -> addressBook.getPersonList().remove(0));
     }
 
+    @Test
+    public void toStringMethod() {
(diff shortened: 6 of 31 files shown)


damithc (changes requested):
Minor typo in commit message subject: `@OverRide` -> `@Override`
The rest LGTM.

pyokagan:
Oh wow, blast from the past :-)

Thanks for keeping my work alive!

pyokagan:
btw if you made any improvements you should absolutely claim some authorship using the `Co-authored-by:` commit message trailer :-)

Eclipse-Dominator:
> Minor typo in commit message subject: `@OverRide` -> `@Override` The rest LGTM.

It's actually not a typo, Github, automatically replaced it as a user mention so it auto corrected it to OverRide.

I will enclose them with code tag to avoid this.

damithc:
> Oh wow, blast from the past :-)
> 
> Thanks for keeping my work alive!

@pyokagan Yes, finally we got around to working on your remaining unmerged PRs at AB4.

damithc:
@se-edu/tech-team-level1 please review this PR if you are around these days ...

src/main/java/seedu/address/commons/util/ToStringBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+     * Constructs a {@code ToStringBuilder} whose formatted output will be prefixed with the
+     * canonical class name of {@code object}.
+     */
+    public ToStringBuilder(Object object) {
+        this(object.getClass().getCanonicalName());
+    }
+
+    /**
+     * Adds a field name/value pair to the output string.
+     */
```
SPWwj: In the add() method, it would be helpful to explain a bit more about the parameters
e.g.
/**
 * Adds a field name/value pair to the output string.
 *
 * @param fieldName The name of the field to be added.
 * @param fieldValue The value of the field to be added.
 * @return A reference to this object, allowing method calls to be chained.
 */

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -10,4 +12,21 @@ public class Messages {
...
+        final StringBuilder builder = new StringBuilder();
+        builder.append(person.getName())
+                .append(" Phone: ")
+                .append(person.getPhone())
+                .append(" Email: ")
+                .append(person.getEmail())
+                .append(" Address: ")
+                .append(person.getAddress())
+                .append(" Tags: ");
+        person.getTags().forEach(builder::append);
```
SPWwj, line 28: Example Output: John Doe Phone: 12345678 Email: johndoe@example.com Address: 123 Street Tags: [friend][classmate]

The Messages class serves as a utility class for formatting and storing user-visible messages. The class is utilized to create a user-friendly, readable string that represents a Person object. However, there are several aspects where this method could be improved:

1. Lack of Attribute Separators: The current formatting of the Person attributes doesn't include clear separators. This may lead to confusion when reading the output, particularly if a Person's name is erroneously interpreted as part of their phone number (for example, "John Doe Phone" might be misinterpreted as "John DoePhone"). Introducing distinct separators between attributes will greatly improve readability. A suggestion would be to include something like a colon (":") or a hyphen ("-") between the attributes.

2. Improving Code Cohesion: As it stands, the Messages class and ToStringBuilder class perform similar tasks, i.e., they both format objects into strings. To improve the code's cohesion, it might be beneficial to merge the Messages class's functionality into the ToStringBuilder class. This can help to keep all the string formatting logic in one place, making it easier for developers to find and use these APIs.

3. Enhancing Readability with ToStringBuilder: Since ToStringBuilder is already used in your codebase, it can be a good choice for implementing the format method, as it's designed for creating readable string representations of objects. The ToStringBuilder could be used to build the string with the added benefit of clear separation of fields, making the output easier to read.

src/test/java/seedu/address/AppParametersTest.java
```diff
@@ -37,6 +37,13 @@ public void parse_invalidConfigPath_success() {
         assertEquals(expected, AppParameters.parse(parametersStub));
     }
 
+    @Test
+    public void toStringMethod() {
+        AppParameters appParameters = new AppParameters();
+        String expected = AppParameters.class.getCanonicalName() + "{configPath=" + appParameters.getConfigPath() + "}";
+        assertEquals(expected, new AppParameters().toString());
+    }
+
```
SPWwj: The provided test suite is designed to test various aspects of an AppParameters class. It appears to be well-structured, with clear and concise tests. However, there are some areas where it could be improved:

1. Naming of the toStringMethod Test Method: Test method names should clearly express what behavior they are verifying. The method toStringMethod could be named more descriptively, such as testToStringProducesExpectedOutput.

2. Use of new AppParameters().toString(): The test creates a new AppParameters object and immediately calls toString(). It would be clearer to reuse the appParameters object, which was created just before, to ensure the expected and actual outputs are based on the same data.

chia-yh (commented):
Should UserPrefs be considered for the uniform toString() format as well? It seems like it fulfils most of the criteria and would benefit from the uniform toString() format.

src/main/java/seedu/address/logic/commands/EditCommand.java
```diff
@@ -120,6 +121,14 @@ public boolean equals(Object other) {
                 && editPersonDescriptor.equals(e.editPersonDescriptor);
     }
 
+    @Override
+    public String toString() {
+        return new ToStringBuilder(this)
+                .add("index", index)
```
cjyothika, line 127: Hi, I think having a uniform toString() format is a great idea!
However, I am not sure if it is expected behaviour for this index to be printed as an index object hash code (eg. seedu.address.commons.core.index.Index@22d9bc14). 
Perhaps the index class requires a toString() method as well if this is not intended.

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -83,6 +83,14 @@ public void getPersonList_modifyList_throwsUnsupportedOperationException() {
         assertThrows(UnsupportedOperationException.class, () -> addressBook.getPersonList().remove(0));
     }
 
+    @Test
+    public void toStringMethod() {
+        String expected = AddressBook.class.getCanonicalName() + "{persons=" + addressBook.getPersonList() + "}";
+        System.out.println(addressBook.toString());
```
SPWwj: You're using System.out.println() to print the results of the toString() method and the expected output. This is likely helpful for your debugging process. However, it's best to remove these lines once your test is working correctly. Test methods ideally should not produce console output, as it can clutter the test result reports.

cjyothika (commented):
The changes look good generally. 

Other than the aforementioned issue with the Index Class, perhaps the returned strings could have more line separators since for classes like EditCommand it is lengthy (eg. seedu.address.logic.commands.EditCommand{index=seedu.address.commons.core.index.Index@22d9bc14, editPersonDescriptor=seedu.address.logic.commands.EditCommand.EditPersonDescriptor{name=null, phone=null, email=null, address=null, tags=null}})

SPWwj (commented):
Awesome job on the pull request!

Here are a few things that jumped out at me:

ToStringBuilder Use: Really liking your use of ToStringBuilder. It's helping us keep our toString methods clean and easy to read.
New Utility Methods: The new utility methods you added, like the format in the Messages class, are super handy. One thing to think about is maybe adding some separators between the attributes to make the output easier to read.
Tests: Your tests are nice and succinct and generally follow good practices. One thing to keep in mind is we could be more descriptive with our test method names and maybe try out some more edge cases.
Code Style: There's some small stuff, like adding spaces around equals signs in toString methods, which could make our code even more readable.
All in all, your PR makes a lot of positive changes. Have a look over the detailed comments for more specific feedback and ideas.

src/main/java/seedu/address/commons/core/Messages.java
Eclipse-Dominator: 2.
I appreciate the suggestion, but I have plans in a future commit/pr to move the `Messages` class to within `logic` package. As it currently stands, all functionalities of `Messages` class is related to displaying messages directly to the user via the GUI and it is right now exclusively used by the `logic` package. Thus, I feel that it is more appropriate to move the `Messages` class directly into the `logic` package instead of keeping it in `commons` package. 

The purpose of `ToStringBuilder` is mainly for devs to easily identify object string representations and their related data in logs and unit testings etc. Thus, it mainly serves as a way to enable easy debugging and testing. To enable this, user-facing messages will thus need to be migrated to use a different formatting method which was what was implemented here. 

3.
I will definitely experiment adding additional purpose/functionalities to `toStringBuilder` based on your suggestion in 3!

damithc: For 1, isn't the code already using `;` as the attribute separator?

damithc:
@Eclipse-Dominator if you updated the code, don't forget to post a new iteration using CIHR

src/main/java/seedu/address/commons/core/Messages.java
SPWwj: > For 1, isn't the code already using `;` as the attribute separator?

The string output from the code when tested is as follows:

"John Doe Phone: 12345678 Email: johndoe@example.com Address: 123 Street Tags: [friend][classmate]"

This output is used for user interface display. The separators between different elements (name, phone, email, address, tags) are crucial for readability and user understanding.

Consider the following examples:

"John Doe Phone" or "Address: 123 Street Tags: [friend][classmate]"

Without proper separators, the information can blend together and become confusing for the user. Therefore, I feel it's important to ensure that the separators are clearly defined and consistently used throughout the output.

damithc: hmm... `toString()` is usually used for internal purposes (such as logging and debugging), to produce a compact string representation of the object. While it _can_ be used for UI purposes, it might not be ideal as UI display have different considerations (easier reading, nicer formatting etc.). Is AB3 using `toString()` method of Person class for UI display?

Eclipse-Dominator: My apologies, I forgot to post a new iteration via CIHR sorry for the confusion, I have manually added the separator `;` in the format.

Regarding adding multiple use case to the `toStringBuilder`, while it is possible I feel that the general use case for it might not as widely use. As mentioned before `toStringBuilder` was originally meant as an easy way to print out object string representation while formatting is related to UI side details. So after some tweaking of code, I feel that there isn't enough usage to justify expanding `toStringBuilder` to allow customizing its display format. 
Perhaps it would be better to reconsider this when there are more use cases in the future which require similar formatting for user side message.

damithc: > hmm... `toString()` is usually used for internal purposes (such as logging and debugging), to produce a compact string representation of the object. While it _can_ be used for UI purposes, it might not be ideal as UI display have different considerations (easier reading, nicer formatting etc.). Is AB3 using `toString()` method of Person class for UI display?

Just realized that this method is specifically if for showring Person object details on the UI. So, the above doesn't apply.

src/test/java/seedu/address/model/person/UniquePersonListTest.java
```diff
@@ -167,4 +167,9 @@ public void asUnmodifiableObservableList_modifyList_throwsUnsupportedOperationEx
         assertThrows(UnsupportedOperationException.class, ()
             -> uniquePersonList.asUnmodifiableObservableList().remove(0));
     }
+
+    @Test
+    public void toStringMethod_producesExpectedOutput() {
```
damithc:
```suggestion
    public void toStringMethod() {
```
In retrospect, this is better. Adding `_producesExpectedOutput` is just stating the obvious. Besides, it does not follow the 3-part naming pattern `methodTested_scenario_expectedBehaivor`
It is fine for the test method to be same as method being tested, unless there are multiple `@Test` methods testing the same method.

damithc (changes requested):
One suggestion about test naming

src/main/java/seedu/address/logic/commands/EditCommand.java
damithc: Good catch @cjyothika 👍

src/main/java/seedu/address/commons/util/ToStringBuilder.java
damithc: Good observation @SPWwj . In general, it is fine to omit `@param`  `@return` etc., from the javadoc if they are obvious from the parameter/method name.

src/test/java/seedu/address/AppParametersTest.java
damithc: @SPWwj Sorry I didn't respond to this earlier. 
1. Yes, it descriptive test names are good to have. In case there is only `@Test` method corresponding to a method being tested, it's better to name it exactly as the method being tested. Giving it a longer name doesn't add more information. Only when there are multiple `@Test` methods that we need to differentiate them using more informative names.
2. Yes, I think so too.

src/test/java/seedu/address/model/AddressBookTest.java
damithc: Good catch @SPWwj

damithc (approved):
LGTM.

chia-yh (commented):
LGTM

src/test/java/seedu/address/logic/commands/FindCommandTest.java
```diff
@@ -74,6 +74,14 @@ public void execute_multipleKeywords_multiplePersonsFound() {
         assertEquals(Arrays.asList(CARL, ELLE, FIONA), model.getFilteredPersonList());
     }
 
+    @Test
+    public void toStringMethod() {
+        NameContainsKeywordsPredicate predicate = new NameContainsKeywordsPredicate(Arrays.asList("keyword"));
+        FindCommand findCommand = new FindCommand(predicate);
+        String expected = FindCommand.class.getCanonicalName() + "{predicate=" + predicate + "}";
```
cjyothika, line 81: I believe the predicate might also benefit from a ToString() method to print the keywords for the FindCommand. Currently, the ToString() method for the FindCommand class prints the hash code of the predicate object it contains.

cjyothika (approved):
Other than the small nitpick pointed out regarding the NameContainsKeywordsPredicate class, I have 2 suggestions that could be considered in a future PR:

1. The output of the ToString() method could be better formatted although for the purpose of internal debugging and testing, the current format should be sufficient. My main gripe with the current format is the line length that could be fixed with line separators being called with the current StringBuilder. However, as suggested by @SPWwj, a separate method for formatting would be a better strategy since it would be less hard coded.
2. I believe that classes like ExitCommand and ListCommand do not have a ToString() method, which I believe is a good decision since there is not much value added from it. However, it does feel inconsistent since objects from these classes will be printed as hash codes instead.

Ultimately, for the purpose of internal debugging and testing, the current iteration LGTM.

SPWwj (approved):
LGTM!

damithc:
> 1. The output of the ToString() method could be better formatted although for the purpose of internal debugging and testing, the current format should be sufficient. My main gripe with the current format is the line length that could be fixed with line separators being called with the current StringBuilder. However, as suggested by @SPWwj, a separate method for formatting would be a better strategy since it would be less hard coded.

@cjyothika Outputs of toString() methods are often combined with outputs of toString() methods of other objects (i.e., when an object is a part of a bigger object structure). Therefore, we cannot assume it will be printed on its own and hence we cannot accurately predict where the best places to put line breaks and how much indentation to use. That is why it is typical for toString to be single liners of format `{key-value pairs}`.

damithc:
Thanks to authors (@pyokagan @Eclipse-Dominator) and the reviewers.

PR #190, opened 2023-06-06
Removal of repeatable parameters

### LINKED ISSUE #162: Give a warning when a parameter is repeated in a user command

Eclipse-Dominator:
This PR an alternative that Fixes #162 (other alternatives: #176, #181)

In this PR, we treat repeated parameters as an error.

```
If a single-valued parameter is repeated multiple times in a 
command, only the last occurrence is taken.  This allows the user
to correct a mistake in a parameter value by simply typing it again,
which may be faster than moving the cursor back to the original
parameter values and editing it. This feature serves as a good
example of 'going the extra mile to optimize the application for
target users' (i.e., users who can type fast).
However, this feature can cause certain user errors to go unnoticed
e.g., unintended repetition caused by a typo in parameter name

Alternatives considered:
(a) Retain current behavior but give a friendly warning to the user
    that a parameter was 'overridden'. 
(b) Remove this feature i.e., treat repeated single-valued
    parameters    as an error, and reject the command with an
    error message.

Alternative (a) requires significantly more complicated code
(see #181), steepening the learning curve for for new programmers.
Given that this application is primarily meant for training
novice programmers, the impact on the learning curve is not worth
the benefit of retaining this feature.

Therefore, let's do option (b).
```

### DIFF
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -57,9 +57,6 @@ AddressBook Level 3 (AB3) is a **desktop app for managing contacts, optimized fo
 * Parameters can be in any order.<br>
   e.g. if the command specifies `n/NAME p/PHONE_NUMBER`, `p/PHONE_NUMBER n/NAME` is also acceptable.
 
-* If a parameter is expected only once in the command but you specified it multiple times, only the last occurrence of the parameter will be taken.<br>
-  e.g. if you specify `p/12341234 p/56785678`, only `p/56785678` will be taken.
-
 * Extraneous parameters for commands that do not take in parameters (such as `help`, `list`, `exit` and `clear`) will be ignored.<br>
   e.g. if the command specifies `help 123`, it will be interpreted as `help`.
 
--- a/src/main/java/seedu/address/logic/Messages.java
+++ b/src/main/java/seedu/address/logic/Messages.java
@@ -1,5 +1,10 @@
 package seedu.address.logic;
 
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import seedu.address.logic.parser.Prefix;
 import seedu.address.model.person.Person;
 
 /**
@@ -11,6 +16,20 @@ public class Messages {
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
     public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+    public static final String MESSAGE_DUPLICATE_FIELDS =
+                "Multiple values specified for the following single-valued field(s): ";
+
+    /**
+     * Returns an error message indicating the duplicate prefixes.
+     */
+    public static String getErrorMessageForDuplicatePrefixes(Prefix... duplicatePrefixes) {
+        assert duplicatePrefixes.length > 0;
+
+        Set<String> duplicateFields =
+                Stream.of(duplicatePrefixes).map(Prefix::toString).collect(Collectors.toSet());
+
+        return MESSAGE_DUPLICATE_FIELDS + String.join(" ", duplicateFields);
+    }
 
     /**
      * Formats the {@code person} for display to the user.
--- a/src/main/java/seedu/address/logic/parser/AddCommandParser.java
+++ b/src/main/java/seedu/address/logic/parser/AddCommandParser.java
@@ -38,6 +38,7 @@ public class AddCommandParser implements Parser<AddCommand> {
             throw new ParseException(String.format(MESSAGE_INVALID_COMMAND_FORMAT, AddCommand.MESSAGE_USAGE));
         }
 
+        argMultimap.verifyNoDuplicatePrefixesFor(PREFIX_NAME, PREFIX_PHONE, PREFIX_EMAIL, PREFIX_ADDRESS);
         Name name = ParserUtil.parseName(argMultimap.getValue(PREFIX_NAME).get());
         Phone phone = ParserUtil.parsePhone(argMultimap.getValue(PREFIX_PHONE).get());
         Email email = ParserUtil.parseEmail(argMultimap.getValue(PREFIX_EMAIL).get());
--- a/src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
+++ b/src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
@@ -5,6 +5,10 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Stream;
+
+import seedu.address.logic.Messages;
+import seedu.address.logic.parser.exceptions.ParseException;
 
 /**
  * Stores mapping of prefixes to their respective arguments.
@@ -56,5 +60,19 @@ public class ArgumentMultimap {
      */
     public String getPreamble() {
         return getValue(new Prefix("")).orElse("");
+    }
+
+    /**
+     * Throws a {@code ParseException} if any of the prefixes given in {@code prefixes} appeared more than
+     * once among the arguments.
+     */
+    public void verifyNoDuplicatePrefixesFor(Prefix... prefixes) throws ParseException {
+        Prefix[] duplicatedPrefixes = Stream.of(prefixes).distinct()
+                .filter(prefix -> argMultimap.containsKey(prefix) && argMultimap.get(prefix).size() > 1)
+                .toArray(Prefix[]::new);
+
+        if (duplicatedPrefixes.length > 0) {
+            throw new ParseException(Messages.getErrorMessageForDuplicatePrefixes(duplicatedPrefixes));
+        }
     }
 }
--- a/src/main/java/seedu/address/logic/parser/EditCommandParser.java
+++ b/src/main/java/seedu/address/logic/parser/EditCommandParser.java
@@ -42,7 +42,10 @@ public class EditCommandParser implements Parser<EditCommand> {
             throw new ParseException(String.format(MESSAGE_INVALID_COMMAND_FORMAT, EditCommand.MESSAGE_USAGE), pe);
         }
 
+        argMultimap.verifyNoDuplicatePrefixesFor(PREFIX_NAME, PREFIX_PHONE, PREFIX_EMAIL, PREFIX_ADDRESS);
+
         EditPersonDescriptor editPersonDescriptor = new EditPersonDescriptor();
+
         if (argMultimap.getValue(PREFIX_NAME).isPresent()) {
             editPersonDescriptor.setName(ParserUtil.parseName(argMultimap.getValue(PREFIX_NAME).get()));
         }
--- a/src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
+++ b/src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
@@ -24,6 +24,10 @@ import static seedu.address.logic.commands.CommandTestUtil.VALID_NAME_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_PHONE_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_FRIEND;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_HUSBAND;
+import static seedu.address.logic.parser.CliSyntax.PREFIX_ADDRESS;
+import static seedu.address.logic.parser.CliSyntax.PREFIX_EMAIL;
+import static seedu.address.logic.parser.CliSyntax.PREFIX_NAME;
+import static seedu.address.logic.parser.CliSyntax.PREFIX_PHONE;
 import static seedu.address.logic.parser.CommandParserTestUtil.assertParseFailure;
 import static seedu.address.logic.parser.CommandParserTestUtil.assertParseSuccess;
 import static seedu.address.testutil.TypicalPersons.AMY;
@@ -31,6 +35,7 @@ import static seedu.address.testutil.TypicalPersons.BOB;
 
 import org.junit.jupiter.api.Test;
 
+import seedu.address.logic.Messages;
 import seedu.address.logic.commands.AddCommand;
 import seedu.address.model.person.Address;
 import seedu.address.model.person.Email;
@@ -51,27 +56,77 @@ public class AddCommandParserTest {
         assertParseSuccess(parser, PREAMBLE_WHITESPACE + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
                 + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
 
-        // multiple names - last name accepted
-        assertParseSuccess(parser, NAME_DESC_AMY + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
-                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
-
-        // multiple phones - last phone accepted
-        assertParseSuccess(parser, NAME_DESC_BOB + PHONE_DESC_AMY + PHONE_DESC_BOB + EMAIL_DESC_BOB
-                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
-
-        // multiple emails - last email accepted
-        assertParseSuccess(parser, NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_AMY + EMAIL_DESC_BOB
-                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
-
-        // multiple addresses - last address accepted
-        assertParseSuccess(parser, NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB + ADDRESS_DESC_AMY
-                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
 
         // multiple tags - all accepted
         Person expectedPersonMultipleTags = new PersonBuilder(BOB).withTags(VALID_TAG_FRIEND, VALID_TAG_HUSBAND)
                 .build();
-        assertParseSuccess(parser, NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB + ADDRESS_DESC_BOB
-                + TAG_DESC_HUSBAND + TAG_DESC_FRIEND, new AddCommand(expectedPersonMultipleTags));
+        assertParseSuccess(parser,
+                NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB + ADDRESS_DESC_BOB + TAG_DESC_HUSBAND + TAG_DESC_FRIEND,
+                new AddCommand(expectedPersonMultipleTags));
+    }
+
+    @Test
+    public void parse_repeatedNonTagValue_failure() {
+        String validExpectedPersonString = NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND;
+
+        // multiple names
+        assertParseFailure(parser, NAME_DESC_AMY + validExpectedPersonString,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_NAME));
+
+        // multiple phones
+        assertParseFailure(parser, PHONE_DESC_AMY + validExpectedPersonString,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_PHONE));
+
+        // multiple emails
+        assertParseFailure(parser, EMAIL_DESC_AMY + validExpectedPersonString,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_EMAIL));
+
+        // multiple addresses
+        assertParseFailure(parser, ADDRESS_DESC_AMY + validExpectedPersonString,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_ADDRESS));
+
+        // multiple fields repeated
+        assertParseFailure(parser,
+                validExpectedPersonString + PHONE_DESC_AMY + EMAIL_DESC_AMY + NAME_DESC_AMY + ADDRESS_DESC_AMY
+                        + validExpectedPersonString,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_NAME, PREFIX_ADDRESS, PREFIX_EMAIL, PREFIX_PHONE));
+
+        // invalid value followed by valid value
+
+        // invalid name
+        assertParseFailure(parser, INVALID_NAME_DESC + validExpectedPersonString,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_NAME));
+
+        // invalid email
+        assertParseFailure(parser, INVALID_EMAIL_DESC + validExpectedPersonString,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_EMAIL));
+
+        // invalid phone
+        assertParseFailure(parser, INVALID_PHONE_DESC + validExpectedPersonString,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_PHONE));
+
+        // invalid address
+        assertParseFailure(parser, INVALID_ADDRESS_DESC + validExpectedPersonString,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_ADDRESS));
+
+        // valid value followed by invalid value
+
+        // invalid name
+        assertParseFailure(parser, validExpectedPersonString + INVALID_NAME_DESC,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_NAME));
+
+        // invalid email
+        assertParseFailure(parser, validExpectedPersonString + INVALID_EMAIL_DESC,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_EMAIL));
+
+        // invalid phone
+        assertParseFailure(parser, validExpectedPersonString + INVALID_PHONE_DESC,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_PHONE));
+
+        // invalid address
+        assertParseFailure(parser, validExpectedPersonString + INVALID_ADDRESS_DESC,
+                Messages.getErrorMessageForDuplicatePrefixes(PREFIX_ADDRESS));
     }
 
     @Test
--- a/src/test/java/seedu/address/logic/parser/EditCommandParserTest.java
+++ b/src/test/java/seedu/address/logic/parser/EditCommandParserTest.java
@@ -16,14 +16,15 @@ import static seedu.address.logic.commands.CommandTestUtil.PHONE_DESC_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.TAG_DESC_FRIEND;
 import static seedu.address.logic.commands.CommandTestUtil.TAG_DESC_HUSBAND;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_ADDRESS_AMY;
-import static seedu.address.logic.commands.CommandTestUtil.VALID_ADDRESS_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_EMAIL_AMY;
-import static seedu.address.logic.commands.CommandTestUtil.VALID_EMAIL_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_NAME_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_PHONE_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_PHONE_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_FRIEND;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_HUSBAND;
+import static seedu.address.logic.parser.CliSyntax.PREFIX_ADDRESS;
+import static seedu.address.logic.parser.CliSyntax.PREFIX_EMAIL;
+import static seedu.address.logic.parser.CliSyntax.PREFIX_PHONE;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_TAG;
 import static seedu.address.logic.parser.CommandParserTestUtil.assertParseFailure;
 import static seedu.address.logic.parser.CommandParserTestUtil.assertParseSuccess;
@@ -34,6 +35,7 @@ import static seedu.address.testutil.TypicalIndexes.INDEX_THIRD_PERSON;
 import org.junit.jupiter.api.Test;
 
 import seedu.address.commons.core.index.Index;
+import seedu.address.logic.Messages;
 import seedu.address.logic.commands.EditCommand;
 import seedu.address.logic.commands.EditCommand.EditPersonDescriptor;
 import seedu.address.model.person.Address;
@@ -89,10 +91,6 @@ public class EditCommandParserTest {
 
         // invalid phone followed by valid email
         assertParseFailure(parser, "1" + INVALID_PHONE_DESC + EMAIL_DESC_AMY, Phone.MESSAGE_CONSTRAINTS);
-
-        // valid phone followed by invalid phone. The test case for invalid phone followed by valid phone
-        // is tested at {@code parse_invalidValueFollowedByValidValue_success()}
-        assertParseFailure(parser, "1" + PHONE_DESC_BOB + INVALID_PHONE_DESC, Phone.MESSAGE_CONSTRAINTS);
 
         // while parsing {@code PREFIX_TAG} alone will reset the tags of the {@code Person} being edited,
         // parsing it together with a valid tag results in error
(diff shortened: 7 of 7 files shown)


damithc:
@Eclipse-Dominator Now that we have three alternatives, what your thoughts about selecting one?
Also, in each of the three PR descriptions, you can mention the other two so that anyone visiting the PR can see how they fit into the big picture.

Eclipse-Dominator:

damithc:
> Repeated Parameters show warning: #176
> Repeated Parameters show warning of repeated prefixes: #181
> Repeated Parametes treated as errors: #190

Updated the [PR description](https://github.com/se-edu/addressbook-level3/pull/190#issue-1744773620) instead.

src/main/java/seedu/address/logic/parser/EditCommandParser.java
```diff
@@ -78,5 +87,4 @@ private Optional<Set<Tag>> parseTagsForEdit(Collection<String> tags) throws Pars
         Collection<String> tagSet = tags.size() == 1 && tags.contains("") ? Collections.emptySet() : tags;
         return Optional.of(ParserUtil.parseTags(tagSet));
     }
-
```
damithc: Avoid unrelated changes

src/main/java/seedu/address/logic/parser/AddCommandParser.java
```diff
@@ -44,6 +45,13 @@ public AddCommand parse(String args) throws ParseException {
         Address address = ParserUtil.parseAddress(argMultimap.getValue(PREFIX_ADDRESS).get());
         Set<Tag> tagList = ParserUtil.parseTags(argMultimap.getAllValues(PREFIX_TAG));
 
+        Set<Prefix> duplicatePrefixes = argMultimap.getDuplicatePrefixes(PREFIX_NAME, PREFIX_PHONE,
+                PREFIX_EMAIL, PREFIX_ADDRESS);
+
+        if (!duplicatePrefixes.isEmpty()) {
+            throw new ParseException(Messages.formatDuplicatePrefixesToMessage(duplicatePrefixes));
+        }
```
damithc: Instead of the above, would it be better to be able do this?
```java
argMultimap.verifyNoDuplicatePrefixesFor(PREFIX_NAME, PREFIX_PHONE, PREFIX_EMAIL, PREFIX_ADDRESS);
```

src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
```diff
@@ -51,27 +58,59 @@ public void parse_allFieldsPresent_success() {
...
+                .withTags(VALID_TAG_FRIEND, VALID_TAG_HUSBAND).build();
+        assertParseSuccess(parser, NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB + ADDRESS_DESC_BOB
+                 + TAG_DESC_HUSBAND + TAG_DESC_FRIEND, new AddCommand(expectedPersonMultipleTags));
+    }
 
-        // multiple emails - last email accepted
-        assertParseSuccess(parser, NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_AMY + EMAIL_DESC_BOB
-                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
+    @Test
+    public void parse_repeatedNonTagValue_success() {
```
damithc: `_success()` or `_exceptionThrown()`?

damithc (commented):
Some initial comments ...

damithc:
@Eclipse-Dominator As discussed, let's proceed with this approach.

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -9,5 +14,22 @@ public class Messages {
...
+     * Formats the duplicate prefixes into an error message.
+     */
+    public static String formatDuplicatePrefixesToMessage(Set<Prefix> duplicatePrefixes) {
+        if (duplicatePrefixes.size() == 0) {
+            return "";
+        }
+
+        Set<String> duplicateFields = duplicatePrefixes.stream().map(Prefix::toString).collect(Collectors.toSet());
+
+
```
damithc: Is double blank lines necessary?

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -9,5 +14,22 @@ public class Messages {
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
     public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+    public static final String MESSAGE_DUPLICATE_FIELDS = "Detected multiple values for the following field%s: "
+            + "%s.\nPlease only a single value for those fields\n";
+
+    /**
+     * Formats the duplicate prefixes into an error message.
+     */
+    public static String formatDuplicatePrefixesToMessage(Set<Prefix> duplicatePrefixes) {
```
damithc: `getErrorMessageForDuplicatePrefixes`?

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -57,4 +59,19 @@ public List<String> getAllValues(Prefix prefix) {
     public String getPreamble() {
         return getValue(new Prefix("")).orElse("");
     }
+
+    /**
+     * Returns a set of all prefixes given in the {@code prefixes} that appeared more than once in
+     * the argument multimap.
+     */
+    public Set<Prefix> getDuplicatePrefixes(Prefix... prefixes) {
+        Set<Prefix> ret = new HashSet<>();
```
damithc: `ret`: use full words unless the short form is widely used.

src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
```diff
@@ -51,27 +58,59 @@ public void parse_allFieldsPresent_success() {
...
+        String validExpectedPerson = NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
```
damithc: variable name suggests this is a person object, but it is not.

src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
```diff
@@ -51,27 +58,59 @@ public void parse_allFieldsPresent_success() {
...
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND;
 
-        // multiple addresses - last address accepted
-        assertParseSuccess(parser, NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB + ADDRESS_DESC_AMY
-                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
+        // multiple fields repeated
```
damithc: It's more typical to test for individual errors before testing for a mix of errors.

damithc (changes requested):
Added a few more comments.

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -9,5 +14,21 @@ public class Messages {
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
     public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+    public static final String MESSAGE_DUPLICATE_FIELDS = "Detected multiple values for the following field%s: "
+            + "%s.\nPlease only a single value for those fields\n";
```
jundatan: I feel that this Message should be worded properly.  Perhaps it will be better to have a message like "Multiple values detected for the following field%s: " + "%s.\nOnly one value is allowed for these fields\n".

jundatan (commented):
Suggestion to change the message for duplicate person.

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -9,5 +14,21 @@ public class Messages {
...
+    public static String getDuplicatePrefixesToMessage(Set<Prefix> duplicatePrefixes) {
+        if (duplicatePrefixes.size() == 0) {
+            return "";
+        }
+
+        Set<String> duplicateFields = duplicatePrefixes.stream().map(Prefix::toString).collect(Collectors.toSet());
+
+        return String.format(MESSAGE_DUPLICATE_FIELDS, duplicateFields.size() > 1 ? "s" : "",
+                String.join(" ", duplicateFields), duplicateFields.size() > 1 ? "these" : "this");
+    }
```
damithc: I think we are complicating this code too much for too little payback. No need to bother with singular/plural differences. Something like the following?

`Multiple values specified for the following single-valued filed(s): ...`

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -57,4 +62,24 @@ public List<String> getAllValues(Prefix prefix) {
     public String getPreamble() {
         return getValue(new Prefix("")).orElse("");
     }
+
+    /**
+     * throws a ParseException if any of the prefixes given in {@code prefixes} appeared more than
```
damithc:
```suggestion
     * Throws a (@code ParseException} if any of the prefixes given in {@code prefixes} appeared more than
```

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -57,4 +62,24 @@ public List<String> getAllValues(Prefix prefix) {
...
+     * once in the argument multimap.
```
damithc:
```suggestion
     * once among the arguments.
```

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -57,4 +62,24 @@ public List<String> getAllValues(Prefix prefix) {
...
+            if (list.size() > 1 && prefixSet.contains(key)) {
+                duplicatedPrefixes.add(key);
+            }
+        });
+
+        if (duplicatedPrefixes.isEmpty()) {
+            return;
+        }
+
+        throw new ParseException(Messages.getDuplicatePrefixesToMessage(duplicatedPrefixes));
```
damithc: alternative:
```suggestion
        if (!duplicatedPrefixes.isEmpty()) {
            throw new ParseException(Messages.getDuplicatePrefixesToMessage(duplicatedPrefixes));
        }
        
```

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -9,5 +14,21 @@ public class Messages {
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
     public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+    public static final String MESSAGE_DUPLICATE_FIELDS = "Detected multiple values for the following field%1$s: "
+            + "%2$s.\nOnly one value is allowed for %3$s field%1$s\n";
+
+    /**
+     * Formats the duplicate prefixes into an error message.
+     */
+    public static String getDuplicatePrefixesToMessage(Set<Prefix> duplicatePrefixes) {
```
damithc: alternative to consider:
```suggestion
    public static String getDuplicatePrefixesToMessage(Prefix... duplicatePrefixes) {
```

damithc (commented):
Some suggestions added

src/main/java/seedu/address/commons/core/Messages.java
Eclipse-Dominator: Personally, I think over complicating this code is okay here since everything related to the messages have been abstracted to within `Messages`. So, there won't be any additional code duplication and its self-contained to within `Messages`.

That being said, I think changing it the message to
```
Multiple values specified for the following single-valued field(s): ...
Please only supply a single value for the field(s) mentioned above.
```
can make it appear more user-friendly?

src/main/java/seedu/address/commons/core/Messages.java
Eclipse-Dominator: I kept it intentionally this to avoid ordering of the fields in the message to avoid complications when doing unit test etc. 
It is possible to sort the field within this method but I feel that supplying it with a `Set` will make it cleaner.

src/main/java/seedu/address/commons/core/Messages.java
damithc: >Personally, I think over complicating this code is okay here since everything related to the messages have been abstracted to within Messages. So, there won't be any additional code duplication and its self-contained to within Messages.

**Always** be as economical as possible with code. Every line of code is an additional burden every project member has to carry for the rest of the lifetime of the project.

>That being said, I think changing it the message to
>
>Multiple values specified for the following single-valued field(s): ...
Please only supply a single value for the field(s) mentioned above.
can make it appear more user-friendly?

Not totally opposed, but I have some concerns:
1. Feels a bit redundant. Check how the UI present this error to the user. Perhaps the first sentence is enough for the user to understand the problem.
2. Check other error messages. If they don't have `Please do such and such ...` part, we don't want to make this specific error message different from others.

src/main/java/seedu/address/commons/core/Messages.java
damithc: Instead of creating `Set` objects everywhere this method is called, you can create the `Set` inside the method at the start?

docs/UserGuide.md
```diff
@@ -57,8 +57,8 @@ AddressBook Level 3 (AB3) is a **desktop app for managing contacts, optimized fo
 * Parameters can be in any order.<br>
   e.g. if the command specifies `n/NAME p/PHONE_NUMBER`, `p/PHONE_NUMBER n/NAME` is also acceptable.
 
-* If a parameter is expected only once in the command but you specified it multiple times, only the last occurrence of the parameter will be taken.<br>
-  e.g. if you specify `p/12341234 p/56785678`, only `p/56785678` will be taken.
+* Parameters that is expected to only appear once in the command will be treated as an invalid command if you specified it multiple times.<br>
+  e.g. if you specify `p/12341234 p/56785678`, the application will treat this as an invalid command.
```
damithc: Good catch.
1. As this was a 'special' behavior, we had to specifically mention it in the UG. But now that we have removed that special behavior, just removing the said documentation is enough. No need to explain the updated behavior, as that is what one would expect anyway.
2. This UG change is tightly related to the behavior change. Should be in the same commit. Each commit should take the code from one 'correct' state to another 'correct' state. A commit that updates the behavior without updating the UG takes the code to an incorrect state.

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -9,5 +14,21 @@ public class Messages {
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
     public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+    public static final String MESSAGE_DUPLICATE_FIELDS =
+            "Multiple values specified for the following single-valued field%1$s: %2$s\n"
+                    + "Please only supply a single value for the field%1$s mentioned above.";
```
damithc: Line wrapping is odd.

src/test/java/seedu/address/logic/parser/EditCommandParserTest.java
```diff
@@ -166,36 +166,37 @@ public void parse_oneFieldSpecified_success() {
     }
 
     @Test
-    public void parse_multipleRepeatedFields_acceptsLast() {
+    public void parse_multipleRepeatedFields_failure() {
```
damithc, line 167: I think I'm the one who suggested to use `_failure()` but I regret that now. We don't expect tests to fail. A better way to indicated the expected results is `_exceptionThrown()`

damithc (changes requested):
Some further suggestions

src/main/java/seedu/address/commons/core/Messages.java
Eclipse-Dominator: Yes, but the code technically is really only called once in production code and the rest of it is called in unit test.

Which is called below.
```java
    /**
     * Throws a {@code ParseException} if any of the prefixes given in {@code prefixes} appeared more than
     * once among the arguments.
     */
    public void verifyNoDuplicatePrefixesFor(Prefix... prefixes) throws ParseException {
        Set<Prefix> duplicatedPrefixes = new HashSet<>();
        Set<Prefix> prefixSet = Set.of(prefixes);
        argMultimap.forEach((key, list) -> {
            if (list.size() > 1 && prefixSet.contains(key)) {
                duplicatedPrefixes.add(key);
            }
        });

        if (!duplicatedPrefixes.isEmpty()) {
            throw new ParseException(Messages.getDuplicatePrefixesToMessage(duplicatedPrefixes));
        }
    }
```
Converting `getDuplicatePrefixesToMessage()` to take in `Prefix...` or `Prefix[]` will involve calling `duplicatedPrefixes.toArray(Prefix[]::new))` which will cause it to be converted once again into a `Set` object when inside the method. 

This is true unless we are using a fixed-size array for `duplicatedPrefixes` or alternative we can convert the whole process as a java stream and convert it to an array. 

I will experiment processing the above with java stream and see its readability in the next version.

Eclipse-Dominator: I think the code is cleaner with java stream, I will update the relevant code.

src/test/java/seedu/address/logic/parser/EditCommandParserTest.java
Eclipse-Dominator: Perhaps we can change `_failure()` to `_exceptionThrown` or some more appropriate terms to standardize the unit test formatting in a separate PR for the other unit tests

damithc: Ah... so it is used elsewhere as well. In that case leave it be, for consistency.

src/main/java/seedu/address/commons/core/Messages.java
damithc: BTW, you are doing something very similar in here:
```java
 public void verifyNoDuplicatePrefixesFor(Prefix... prefixes) throws ParseException {
        Set<Prefix> duplicatedPrefixes = new HashSet<>();
        Set<Prefix> prefixSet = Set.of(prefixes);
```

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -9,5 +15,19 @@ public class Messages {
...
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+    public static final String MESSAGE_DUPLICATE_FIELDS =
+            "Multiple values specified for the following single-valued field(s): %s";
+    /**
+     * Formats the duplicate prefixes into an error message.
+     */
+    public static String getDuplicatePrefixesToMessage(Prefix... duplicatePrefixes) {
+        if (duplicatePrefixes.length == 0) {
+            return "";
+        }
```
damithc: This should not even happen, right?
In that case, use an assert instead.

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -9,5 +15,19 @@ public class Messages {
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
     public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+    public static final String MESSAGE_DUPLICATE_FIELDS =
+            "Multiple values specified for the following single-valued field(s): %s";
```
damithc: There should be a blank line here?

damithc: Do we need the `%s`?

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -9,5 +15,19 @@ public class Messages {
...
+    /**
+     * Formats the duplicate prefixes into an error message.
+     */
+    public static String getDuplicatePrefixesToMessage(Prefix... duplicatePrefixes) {
+        if (duplicatePrefixes.length == 0) {
+            return "";
+        }
+
+        Set<String> duplicateFields =
+                Stream.of(duplicatePrefixes).map(Prefix::toString).collect(Collectors.toSet());
```
damithc: Our policy is to line-wrap only when close to the line limit, right? e.g.,
```suggestion
        Set<String> duplicateFields = Stream.of(duplicatePrefixes).map(Prefix::toString)
                .collect(Collectors.toSet());
```

Follow that policy consistently.

src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
```diff
@@ -51,27 +56,76 @@ public void parse_allFieldsPresent_success() {
...
+
+        // invalid address
+        assertParseFailure(parser, validExpectedPersonString + INVALID_ADDRESS_DESC,
+                Messages.getDuplicatePrefixesToMessage(PREFIX_ADDRESS));
+
+        // multiple fields repeated
+        assertParseFailure(parser,
+                validExpectedPersonString + PHONE_DESC_AMY + EMAIL_DESC_AMY + NAME_DESC_AMY
+                        + ADDRESS_DESC_AMY + validExpectedPersonString,
+                Messages.getDuplicatePrefixesToMessage(PREFIX_NAME, PREFIX_ADDRESS, PREFIX_EMAIL, PREFIX_PHONE));
```
damithc: This can be moved just above `// invalid value followed by valid value`. Otherwise it appears as if belonging to `// valid value followed by invalid value` category.

src/test/java/seedu/address/logic/parser/EditCommandParserTest.java
```diff
@@ -90,10 +92,6 @@ public void parse_invalidValue_failure() {
         // invalid phone followed by valid email
         assertParseFailure(parser, "1" + INVALID_PHONE_DESC + EMAIL_DESC_AMY, Phone.MESSAGE_CONSTRAINTS);
 
-        // valid phone followed by invalid phone. The test case for invalid phone followed by valid phone
-        // is tested at {@code parse_invalidValueFollowedByValidValue_success()}
-        assertParseFailure(parser, "1" + PHONE_DESC_BOB + INVALID_PHONE_DESC, Phone.MESSAGE_CONSTRAINTS);
-
```
damithc, line 96: Good catch.

src/test/java/seedu/address/logic/parser/EditCommandParserTest.java
```diff
@@ -166,36 +164,34 @@ public void parse_oneFieldSpecified_success() {
...
 
-        // other valid values specified
-        userInput = targetIndex.getOneBased() + EMAIL_DESC_BOB + INVALID_PHONE_DESC + ADDRESS_DESC_BOB
-                + PHONE_DESC_BOB;
-        descriptor = new EditPersonDescriptorBuilder().withPhone(VALID_PHONE_BOB).withEmail(VALID_EMAIL_BOB)
-                .withAddress(VALID_ADDRESS_BOB).build();
-        expectedCommand = new EditCommand(targetIndex, descriptor);
-        assertParseSuccess(parser, userInput, expectedCommand);
+        assertParseFailure(parser, userInput,
+                Messages.getDuplicatePrefixesToMessage(PREFIX_PHONE, PREFIX_EMAIL, PREFIX_ADDRESS));
```
damithc: It seems you do more extensive testing of this aspect for the Add command than the Edit command. Perhaps we should make a note of that somewhere in this class? e.g.,
`/* More extensive testing of duplicate parameter detections is done in ...*/`

damithc (changes requested):
Some suggestions added (mostly cosmetic)

src/test/java/seedu/address/logic/parser/EditCommandParserTest.java
```diff
@@ -166,36 +164,39 @@ public void parse_oneFieldSpecified_success() {
...
+         * AddCommandParserTest#parse_repeatedNonTagValue_failure()
+         */
+
+        // valid followed by invalid
         Index targetIndex = INDEX_FIRST_PERSON;
-        String userInput = targetIndex.getOneBased() + PHONE_DESC_AMY + ADDRESS_DESC_AMY + EMAIL_DESC_AMY
+        String userInput = targetIndex.getOneBased() + INVALID_PHONE_DESC + PHONE_DESC_BOB;
+
+        assertParseFailure(parser, userInput,
+                        Messages.getDuplicatePrefixesToMessage(PREFIX_PHONE));
```
damithc: Unnecessary line break?

src/test/java/seedu/address/logic/parser/EditCommandParserTest.java
```diff
@@ -166,36 +164,39 @@ public void parse_oneFieldSpecified_success() {
...
+
+        // invalid followed by valid
+        userInput = targetIndex.getOneBased() + PHONE_DESC_BOB + INVALID_PHONE_DESC;
+
+        assertParseFailure(parser, userInput,
+                        Messages.getDuplicatePrefixesToMessage(PREFIX_PHONE));
```
damithc: Same here

docs/UserGuide.md
```diff
@@ -57,9 +57,6 @@ AddressBook Level 3 (AB3) is a **desktop app for managing contacts, optimized fo
 * Parameters can be in any order.<br>
```
damithc, line 57: Added a few cosmetic suggestions.

Commit message:
```
Currently, the behavior of repeated parameters in AddCommand and
EditCommand is to consider the last value used. This behavior can be
confusing and difficult to work with. Therefore, instead of allowing
repeatable parameters, it is better to present them as a Command Error
to the user.
```
Has vague parts e.g., 'difficult to work with' (why is it difficult to work with?)
You can refine this based on the description given in https://github.com/se-edu/addressbook-level3/issues/162


```
To implement this change, the checking for duplicate parameters should
be performed before checking the validity of individual fields. In other
words, if an Add or Edit Command contains multiple duplicate fields that
are all invalid, the parsing error for duplicate fields will take precedence
over other checks for individual parameters.
```
* What is the justification for choosing this behavior? Commit messages should cover the WHY aspect. 
* This explanation doesn't cover cases of multiple values only _some_ of which are invalid. It only mentions cases where _all_ of them are invalid.

Eclipse-Dominator:
Resolve rebase conflicts.

damithc:
@Eclipse-Dominator Looks like the rebase wasn't entirely correct. There are some unrelated code in this PR.

docs/UserGuide.md
damithc: I added these comments sometime back but it was stuck in 'draft' status.

docs/UserGuide.md
```diff
@@ -57,9 +57,6 @@ AddressBook Level 3 (AB3) is a **desktop app for managing contacts, optimized fo
...
   e.g. if the command specifies `n/NAME p/PHONE_NUMBER`, `p/PHONE_NUMBER n/NAME` is also acceptable.
```
damithc, line 58: Suggested tweaks to the commit message:

```
Remove support for repeated parameters in commands

If a single-valued parameter is repeated multiple times in a 
command, only the last occurrence is taken.  This allows the user
to correct a mistake in a parameter value by simply typing it again,
which may be faster than moving the cursor back to the original
parameter values and editing it. This feature serves as a good
example of 'going the extra mile to optimize the application for
target users' (i.e., users who can type fast).
However, this feature can cause certain user errors to go unnoticed
e.g., unintended repetition caused by a typo in parameter name

Alternatives considered:
(a) Retain current behavior but give a friendly warning to the user
    that a parameter was 'overridden'. 
(b) Remove this feature i.e., treat repeated single-valued
    parameters    as an error, and reject the command with an
    error message.

Alternative (a) requires significantly more complicated code
(see #181), steepening the learning curve for for new programmers.
Given that this application is primarily meant for training
novice programmers, the impact on the learning curve is not worth
the benefit of retaining this feature.

Therefore, let's do option (b).
```

Changes: 
1. Use imperative mood in subject
2. Added more details about the alternatives and the decision

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -11,6 +16,20 @@ public class Messages {
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
     public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+    public static final String MESSAGE_DUPLICATE_FIELDS =
+                "Multiple values specified for the following single-valued field(s): ";
+
+    /**
+     * Formats the duplicate prefixes into an error message.
+     */
+    public static String getDuplicatePrefixesToMessage(Prefix... duplicatePrefixes) {
```
damithc:
```suggestion
    public static String getErrorMessageForDuplicatePrefixes(Prefix... duplicatePrefixes) {
```

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -57,4 +61,18 @@ public List<String> getAllValues(Prefix prefix) {
...
     }
+
+    /**
+     * Throws a {@code ParseException} if any of the prefixes given in {@code prefixes} appeared more than
+     * once among the arguments.
+     */
+    public void verifyNoDuplicatePrefixesFor(Prefix... prefixes) throws ParseException {
+        Prefix[] duplicatedPrefixes = Stream.of(prefixes).distinct().filter(
+                prefix -> argMultimap.containsKey(prefix) && argMultimap.get(prefix).size() > 1)
+                .toArray(Prefix[]::new);
```
damithc: Is this indentation correct?  Alternative:
```suggestion
        Prefix[] duplicatedPrefixes = Stream.of(prefixes).distinct()
                .filter(prefix -> argMultimap.containsKey(prefix) && argMultimap.get(prefix).size() > 1)
                .toArray(Prefix[]::new);
```

src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -1,5 +1,10 @@
 package seedu.address.commons.core;
 
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import seedu.address.logic.parser.Prefix;
 import seedu.address.model.person.Person;
```
damithc: Dependency on `model` is already a problem, and we are supposed to fix that separately. Now we are adding a dependency on `logic`. Would that be a further problem?

damithc (changes requested):
Some comments added (did not review the test code though).

src/main/java/seedu/address/commons/core/Messages.java
damithc: After another look, I think this should be fine, as we plan to move Messages to the `logic` package.

src/test/java/seedu/address/logic/parser/EditCommandParserTest.java
```diff
@@ -166,36 +164,37 @@ public void parse_oneFieldSpecified_success() {
     }
 
     @Test
-    public void parse_multipleRepeatedFields_acceptsLast() {
+    public void parse_multipleRepeatedFields_failure() {
+        /*
+         * More extensive testing of duplicate parameter detections is done in
+         * AddCommandParserTest#parse_repeatedNonTagValue_failure()
+         */
```
damithc: Should we use `//` here?

damithc (commented):
Just a minor question. I think we are close to home on this one. Let me ask others to have a look as well.

damithc:

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -57,4 +61,18 @@ public List<String> getAllValues(Prefix prefix) {
     public String getPreamble() {
         return getValue(new Prefix("")).orElse("");
     }
+
+    /**
+     * Throws a {@code ParseException} if any of the prefixes given in {@code prefixes} appeared more than
```
dlimyy, line 66: For the javadocs, I believe the `@param` and `@throws` tags have to be added
```suggestion
    /**
     * Throws a {@code ParseException} if any of the prefixes given in {@code prefixes} appeared more than
     * once among the arguments.
     *
     * @param prefixes List of prefixes supplied by the user
     * @throws ParseException If duplicate prefixes are present
     */
```

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -57,4 +61,18 @@ public List<String> getAllValues(Prefix prefix) {
...
+     * once among the arguments.
+     */
+    public void verifyNoDuplicatePrefixesFor(Prefix... prefixes) throws ParseException {
```
dlimyy, line 69: Could consider shortening the name of the method to verifyNoDuplicatePrefixes as the 'For' does not seem to be very necessary

src/test/java/seedu/address/logic/parser/EditCommandParserTest.java
```diff
@@ -166,36 +164,37 @@ public void parse_oneFieldSpecified_success() {
...
+        String userInput = targetIndex.getOneBased() + INVALID_PHONE_DESC + PHONE_DESC_BOB;
+
+        assertParseFailure(parser, userInput, Messages.getErrorMessageForDuplicatePrefixes(PREFIX_PHONE));
+
+        // invalid followed by valid
+        userInput = targetIndex.getOneBased() + PHONE_DESC_BOB + INVALID_PHONE_DESC;
+
+        assertParseFailure(parser, userInput, Messages.getErrorMessageForDuplicatePrefixes(PREFIX_PHONE));
+
+        // mulltiple valid field repeated
```
dlimyy: Some minor grammatical errors here. It should be `multiple valid fields repeated` instead.

dlimyy (commented):
Overall, LGTM! Just some minor suggestions below

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
damithc: It reads as a natural statement e.g., `verifyNoDuplicatePrefixesFor(PREFIX_NAME, PREFIX_ADDRESS)` i.e., verify no duplicate prefixes for A, B, ..
But OK to remove. 
Can also consider `verifyNoDuplicatesFor` as Prefix is implied by the parameter type.

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
damithc: @dlimyy Good catch that some tags are missing. Our policy is javadoc tags can be omitted if not adding value e.g., repeating the same information already available elsewhere in the javadoc

docs/UserGuide.md
```diff
@@ -57,9 +57,6 @@ AddressBook Level 3 (AB3) is a **desktop app for managing contacts, optimized fo
...
 
-* If a parameter is expected only once in the command but you specified it multiple times, only the last occurrence of the parameter will be taken.<br>
-  e.g. if you specify `p/12341234 p/56785678`, only `p/56785678` will be taken.
-
 * Extraneous parameters for commands that do not take in parameters (such as `help`, `list`, `exit` and `clear`) will be ignored.<br>
   e.g. if the command specifies `help 123`, it will be interpreted as `help`.
```
SPWwj, line 61:
```suggestion
* Extra parameters will be ignored.
  e.g. if the command is help 123`, it will be interpreted as help. This is also the case for `list`, `exit`, and `clear`.
```

src/main/java/seedu/address/logic/Messages.java
```diff
@@ -11,6 +16,20 @@ public class Messages {
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
     public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
```
SPWwj, line 17:
```suggestion
    public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "Invalid person index";
```

src/main/java/seedu/address/logic/Messages.java
```diff
@@ -11,6 +16,20 @@ public class Messages {
...
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+    public static final String MESSAGE_DUPLICATE_FIELDS =
+                "Multiple values specified for the following single-valued field(s): ";
+
+    /**
+     * Formats the duplicate prefixes into an error message.
```
SPWwj:
```suggestion
     * Generates a formatted error message indicating the duplicate prefixes.
```

SPWwj (changes requested):

docs/UserGuide.md
damithc: Not related to this PR?

src/main/java/seedu/address/logic/Messages.java
damithc: `a formatted error message` -> `an error message`?

src/main/java/seedu/address/logic/Messages.java
Eclipse-Dominator: I think this is not related to the current PR, I think we can fine tune this messages in another PR.

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -57,4 +61,22 @@ public List<String> getAllValues(Prefix prefix) {
...
         return getValue(new Prefix("")).orElse("");
     }
+
+    /**
+     * Throws a {@code ParseException} if any of the prefixes given in {@code prefixes} appeared more than
+     * once among the arguments.
+     *
+     * @param prefixes Prefixes to check for duplicates
+     * @throws ParseException If any of the prefixes given in {@code prefixes} appeared more than once among the
+     *                       arguments
```
damithc: No need, as I mentioned in my comment.

src/main/java/seedu/address/logic/Messages.java
```diff
@@ -11,6 +16,20 @@ public class Messages {
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
     public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
     public static final String MESSAGE_PERSONS_LISTED_OVERVIEW = "%1$d persons listed!";
+    public static final String MESSAGE_DUPLICATE_FIELDS =
+                "Multiple values specified for the following single-valued field(s): ";
+
+    /**
+     * Generates a formatted error message indicating the duplicate prefixes.
```
damithc: `a formatted error message` -> `an error message` (as I mentioned in my previous comment)
While you are at it, `Returns`  may be better than `Generates`.

damithc (changes requested):
Two commented you missed.

SPWwj (approved):
Code works as expected. LGTM.

damithc (approved):

PR #192, opened 2023-06-14
[#98] Space character dropped at line break when converting to PDF 

### LINKED ISSUE #98: Space character dropped at line break when converting to PDF
When converting the UG to PDF format, the space character at a line break can be lost e.g.,

image

This affects users copy-pasting commands from the UG to the app.

Eclipse-Dominator:
Fixes #98 

In the pdf version of UG, commands that span across more than one line when copied-pasted to the application will drop the line-break. As line-breaks will usually be formatted at spaces when converting to pdf, this will cause the space in command to be dropped when pasting into the application.

Therefore, let's update the UG to inform the users about this behavior.

### DIFF
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -63,6 +63,7 @@ AddressBook Level 3 (AB3) is a **desktop app for managing contacts, optimized fo
 * Extraneous parameters for commands that do not take in parameters (such as `help`, `list`, `exit` and `clear`) will be ignored.<br>
   e.g. if the command specifies `help 123`, it will be interpreted as `help`.
 
+* If you are using a PDF version of this document, be careful when copying and pasting commands that span multiple lines as space characters surrounding line-breaks may be omitted when copied over to the application.
 </div>
 
 ### Viewing help : `help`


docs/UserGuide.md
```diff
@@ -63,6 +63,7 @@ AddressBook Level 3 (AB3) is a **desktop app for managing contacts, optimized fo
 * Extraneous parameters for commands that do not take in parameters (such as `help`, `list`, `exit` and `clear`) will be ignored.<br>
   e.g. if the command specifies `help 123`, it will be interpreted as `help`.
 
+* Be careful when copy-pasting commands that cut across a single line from this user guide as `Sample commands` that span more than a single line may drop a space at the line-break when copied over to the application.
```
dlimyy: Can consider shortening `Sample commands ...` to `these commands` to make it less verbose

```suggestion
* Be careful when copy-pasting commands that span more than a single line from this user guide as the space character at the line-break may be lost when these commands are copied over to the application. 
```

dlimyy (commented):
Overall, LGTM! Just some suggestions below

docs/UserGuide.md
damithc: Another alternative that mentions PDF files specifically.
```suggestion
* If you are using a PDF version of this document, be careful when copy-pasting commands that span multiple lines, as spaces surrounding line breaks can get omitted when copy-pasted from a PDF file. 
```

damithc (changes requested):
commit message: 
*span across more than one line -> span multiple lines
* Try using ChatGPT to see if the commit message phrasing can be improved.

UserGuide.md: there is a typo

damithc:
Remember to review commit-by-commit, and review the commit message as well.

docs/UserGuide.md
```diff
@@ -63,6 +63,7 @@ AddressBook Level 3 (AB3) is a **desktop app for managing contacts, optimized fo
 * Extraneous parameters for commands that do not take in parameters (such as `help`, `list`, `exit` and `clear`) will be ignored.<br>
   e.g. if the command specifies `help 123`, it will be interpreted as `help`.
 
+* If you are using a PDF version of this document, be careful when copy-pasting commands that span multiple lines as space characters surrounding line-breaks may be omitted when copied over to the application.
```
jundatan: Imo, is quite odd to put this under the **"notes about the command format"** section as it does not fit well with the other pointers in it. Perhaps, it will be better to be put it in a separate section like the FAQ or below the features header outside of the **"notes about the command format"** section?

SPWwj (commented):
image

The addition of a newline has resulted in the 'help' command being pushed to a new page in the command table, which unfortunately led to an overlap in the rendering. This issue needs to be addressed before proceeding with the merge.

By the way, using incognito mode might be helpful in debugging ;)

The commit message could be more precise and clearer.

Eclipse-Dominator:
> image
> The addition of a newline has resulted in the 'help' command being pushed to a new page in the command table, which unfortunately led to an overlap in the rendering. This issue needs to be addressed before proceeding with the merge.
> 
> By the way, using incognito mode might be helpful in debugging ;)
> 
> The commit message could be more precise and clearer.

@SPWwj, I'm not sure how to get page like rendering when viewing user guide in browser.

But I wasn't actually able to replicate this when converting to PDF with both Github Pages and local jekyll server.
Conversion to pdf tested on:
- Edge 114.0.1823.43 (Official build) (64-bit)
- Chrome 114.0.5735.110 (Official Build) (64-bit)
- Chrome 114.0.5735.134 (Official Build) (64-bit)

image

Update
---

I was able to replicate this issue. It seems to be related to the service that is used to convert the html to pdf.
If the option "Save as PDF" is selected, the table rendering will be normal like above.

However, the rendering issue seems to be caused by Microsoft's print to PDF service.

image

SPWwj:
> > image
> > The addition of a newline has resulted in the 'help' command being pushed to a new page in the command table, which unfortunately led to an overlap in the rendering. This issue needs to be addressed before proceeding with the merge.
> > By the way, using incognito mode might be helpful in debugging ;)
> > The commit message could be more precise and clearer.
> 
> @SPWwj, I'm not sure how to get page like rendering when viewing user guide in browser.
> 
> But I wasn't actually able to replicate this when converting to PDF with both Github Pages and local jekyll server. Conversion to pdf tested on:
> 
> * Edge 114.0.1823.43 (Official build) (64-bit)
> * Chrome 114.0.5735.110 (Official Build) (64-bit)
> * Chrome 114.0.5735.134 (Official Build) (64-bit)
> 
> image
> 
> image

Window:
- Edge Version 114.0.1823.43 (Official build) (64-bit)
- Chrome Version 114.0.5735.134 (Official Build) (64-bit)
The overlapping occur when paper size is set to Letter:
image

My bad, since the recommended setting is A4, this is not a issue...

docs/UserGuide.md
SPWwj: A minor alteration suggestion:
It may be less confusing to users if we adjust the wording regarding the copy-pasting commands. Instead of suggesting that AB3 has a copy-paste command, we could articulate it as 'copying and pasting the command from the PDF version of the User Guide.'

SPWwj: In my opinion, positioning this information here is beneficial as it pertains directly to the use of commands. From personal experience, I don't typically read through the entire User Guide before starting to use the commands. Therefore, having notes about potential pitfalls or issues within the command format section can be extremely useful.

Furthermore, I view the FAQ as an additional, somewhat less critical feature, which should not interfere with or impede the fundamental understanding and usage of the commands. As such, I believe it's acceptable to prioritize the 'Note about the command' section over the FAQ.

SPWwj:
> * Update ug regarding copying commands

Just a minor suggestion for the commit message: "Update UG with clarified instructions on copying commands". While the current commit message is effective, someone without any prior knowledge of the issue might question what 'regarding copying commands' refers to, as it is rather vague. Therefore, if it fits within the character limit, we should aim to enhance its conciseness.

Other than that this PR look good to me :)

docs/UserGuide.md
damithc: Thanks for inputs @jundatan @SPWwj 
Yes it is a bit odd to have in under 'notes about the command format'. Just below it as a separate box is one viable option. Putting it in the FAQ is not a good idea, as @SPWwj mentioned. For now, let's keep it in the current place as it is less disruptive to the flow of the document.

SPWwj (approved):
LGTM!

damithc:
Merged with minor changes to the commit message (revised the subject, removed 'Therefore,').

Thanks for the PR, @Eclipse-Dominator and for the reviews @SPWwj @jundatan @dlimyy

PR #196, opened 2023-06-17
Fix not visible help menu / bump javafx version (to ver 15)

### LINKED ISSUE #193: Window/stage not showing up in KDE
When no minimum window size is set, launching the window will cause it be not visible when launched from linux distros using KDE as desktop manager.

As the minimum size of the `MainWindow` is set before launch, this issue is not apparent. However, for other popups like the help window, this will become apparent as seen below:

image
image

This seems to be an issue that is caused by using KDE Plasma graphical desktop environment which is one of the popular graphical desktop environment.

Tested on:
- Archlinux with KDE Plasma
- Manjaro KDE

Related issue can be found in the following bug reports here: [JDK-8179073](https://bugs.openjdk.org/browse/JDK-8179073) and [JDK-8198761](https://bugs.openjdk.org/browse/JDK-8198761)

Both of this issues occured on KDE desktop which is what lead to the assumption for the cause of the issue.

---
While both issues are marked as unresolved, based on manual testing, bumping Javafx version to >=13 will fix this issue.

However, the next LTS version is ver 17 which may require adding modules to the application.

image

--- 
Another alternative solution is to add a minimum window size via code which is undesirable since the help popup is set to non-resizable which means the viewing experience for the end user may vary based on their system setting.

Eclipse-Dominator:
Fixes #193 

Launching the application on Linux distributions using KDE as the desktop manager resulted in the window not being visible when no minimum window size was set. The problem seems to be specific to the KDE Plasma graphical desktop environment. This has caused the help menu to be not visible on such platforms.

Manual testing indicates that bumping the JavaFX version to 13 or higher resolves the issue. Since version 16 will log a warning with the current project configuration, let's bump the JavaFX version to 15.

---

However, ver 15 is not LTS anymore and ver 11 is approaching its EOL. The next LTS version is ver 17 which will log a warning for not using a modular application.

There was some investigation into converting the project to modular version here but it's not very effective/useful. ([summary](https://github.com/se-edu/addressbook-level3/issues/108#issuecomment-1595777510))

---

You can find an alternative PR that explores using a log to inform the user to ignore the warning message in #199

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ task coverage(type: JacocoReport) {
 
 dependencies {
     String jUnitVersion = '5.4.0'
-    String javaFxVersion = '11.0.2'
+    String javaFxVersion = '15.0.1'
 
     implementation group: 'org.openjfx', name: 'javafx-base', version: javaFxVersion, classifier: 'win'
     implementation group: 'org.openjfx', name: 'javafx-base', version: javaFxVersion, classifier: 'mac'


Eclipse-Dominator:
As mentioned before, JavaFX 15 is still an EOL version. Since JavaFX 11 is reaching its EOL in September, the next version to upgrade to is actually JavaFX 17. As it stands there is no compatibility error with JavaFX 17. It's just that after JavaFX 16, there is now a warning that will be thrown if JavaFX packages are loaded from classpath. [ver-16](https://github.com/openjdk/jfx/blob/master/doc-files/release-notes-16.md)

This is just a warning that does not affect current usage. I have experimented with migrating over to use modules and concluded as we currently stand it does not actually fix the warning issue as in the release version, `module-info` is actually being ignored when packaging into a FAT JAR. This defeats the purpose of moving to modules since the warning will still appear in the release version. 

Ideally, JavaFX should be packaged via JLink which will package the app into an actual executable (removing the need for dependency on Java) but it will become platform dependant. 

---



Even so, it might still be better to consider JavaFX 17 as it is nowhere near its EOL as compared to 11 or 15 if we can ignore the warning.

damithc:
> Ideally, JavaFX should be packaged via JLink which will package the app into an actual executable (removing the need for dependency on Java) but it will become platform dependant.

Probably not tenable, as we can't be distributing multiple versions of the product for students to test during the practical exam. It's hard enough to distribute one .jar file. Nevertheless, it's good to explore this option just to see how it looks like.


> Even so, it might still be better to consider JavaFX 17 as it is nowhere near its EOL as compared to 11 or 15 if we can ignore the warning.

Yes, we can try this. The UG can be updated to ask users to ignore the warning.

Eclipse-Dominator:
Closed in favor of #199 for now

PR #197, opened 2023-06-18
Standardize #equals(Object) formatting

Eclipse-Dominator:
Partially adapted from [AB4#973](https://github.com/se-edu/addressbook-level4/pull/973/)

To go along with some of the equality checks, some methods and constructors are reevaluated to follow defensive programming by adding suitable null checks. 

AB3 has all sorts of #equals(Object) flavors. Some of them are formatted
with nested logic comparison while some contain unnecessary code. It will
enable better readability for others if they were to be standardized to
follow a similar format.

Generally, #equals(Object) should:
* Name the parameter 'other' of the method
* Check if other == this
* Check if other is an instance of this class + comment for null check
* Set a temporary variable otherCLASSNAME
* Perform other related internal checks.

Should the internal variable be nullable, we should use
Objects.equals(var, var2) to compare between the 2.

Doing the above should improve the clarity and ease of
use of #equals(Object) methods.

### DIFF
--- a/docs/tutorials/AddRemark.md
+++ b/docs/tutorials/AddRemark.md
@@ -118,7 +118,6 @@ public class RemarkCommand extends Command {
 
     @Override
     public boolean equals(Object other) {
-        // short circuit if same object
         if (other == this) {
             return true;
         }
@@ -128,7 +127,6 @@ public class RemarkCommand extends Command {
             return false;
         }
 
-        // state check
         RemarkCommand e = (RemarkCommand) other;
         return index.equals(e.index)
                 && remark.equals(e.remark);
--- a/src/main/java/seedu/address/AppParameters.java
+++ b/src/main/java/seedu/address/AppParameters.java
@@ -50,12 +50,13 @@ public class AppParameters {
             return true;
         }
 
+        // instanceof handles nulls
         if (!(other instanceof AppParameters)) {
             return false;
         }
 
         AppParameters otherAppParameters = (AppParameters) other;
-        return Objects.equals(getConfigPath(), otherAppParameters.getConfigPath());
+        return Objects.equals(configPath, otherAppParameters.configPath);
     }
 
     @Override
--- a/src/main/java/seedu/address/commons/core/Config.java
+++ b/src/main/java/seedu/address/commons/core/Config.java
@@ -39,14 +39,15 @@ public class Config {
         if (other == this) {
             return true;
         }
-        if (!(other instanceof Config)) { //this handles null as well.
+
+        // instanceof handles nulls
+        if (!(other instanceof Config)) {
             return false;
         }
 
-        Config o = (Config) other;
-
-        return Objects.equals(logLevel, o.logLevel)
-                && Objects.equals(userPrefsFilePath, o.userPrefsFilePath);
+        Config otherConfig = (Config) other;
+        return Objects.equals(logLevel, otherConfig.logLevel)
+                && Objects.equals(userPrefsFilePath, otherConfig.userPrefsFilePath);
     }
 
     @Override
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -11,6 +11,7 @@ import static seedu.address.model.Model.PREDICATE_SHOW_ALL_PERSONS;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 
@@ -105,7 +106,6 @@ public class EditCommand extends Command {
 
     @Override
     public boolean equals(Object other) {
-        // short circuit if same object
         if (other == this) {
             return true;
         }
@@ -115,10 +115,9 @@ public class EditCommand extends Command {
             return false;
         }
 
-        // state check
-        EditCommand e = (EditCommand) other;
-        return index.equals(e.index)
-                && editPersonDescriptor.equals(e.editPersonDescriptor);
+        EditCommand otherEditCommand = (EditCommand) other;
+        return index.equals(otherEditCommand.index)
+                && editPersonDescriptor.equals(otherEditCommand.editPersonDescriptor);
     }
 
     @Override
@@ -212,7 +211,6 @@ public class EditCommand extends Command {
 
         @Override
         public boolean equals(Object other) {
-            // short circuit if same object
             if (other == this) {
                 return true;
             }
@@ -222,14 +220,12 @@ public class EditCommand extends Command {
                 return false;
             }
 
-            // state check
-            EditPersonDescriptor e = (EditPersonDescriptor) other;
-
-            return getName().equals(e.getName())
-                    && getPhone().equals(e.getPhone())
-                    && getEmail().equals(e.getEmail())
-                    && getAddress().equals(e.getAddress())
-                    && getTags().equals(e.getTags());
+            EditPersonDescriptor otherEditPersonDescriptor = (EditPersonDescriptor) other;
+            return Objects.equals(name, otherEditPersonDescriptor.name)
+                    && Objects.equals(phone, otherEditPersonDescriptor.phone)
+                    && Objects.equals(email, otherEditPersonDescriptor.email)
+                    && Objects.equals(address, otherEditPersonDescriptor.address)
+                    && Objects.equals(tags, otherEditPersonDescriptor.tags);
         }
 
         @Override
--- a/src/main/java/seedu/address/model/person/NameContainsKeywordsPredicate.java
+++ b/src/main/java/seedu/address/model/person/NameContainsKeywordsPredicate.java
@@ -24,9 +24,17 @@ public class NameContainsKeywordsPredicate implements Predicate<Person> {
 
     @Override
     public boolean equals(Object other) {
-        return other == this // short circuit if same object
-                || (other instanceof NameContainsKeywordsPredicate // instanceof handles nulls
-                && keywords.equals(((NameContainsKeywordsPredicate) other).keywords)); // state check
+        if (other == this) {
+            return true;
+        }
+
+        // instanceof handles nulls
+        if (!(other instanceof NameContainsKeywordsPredicate)) {
+            return false;
+        }
+
+        NameContainsKeywordsPredicate otherNameContainsKeywordsPredicate = (NameContainsKeywordsPredicate) other;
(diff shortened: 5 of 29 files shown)


src/main/java/seedu/address/commons/util/ToStringBuilder.java
```diff
@@ -1,5 +1,7 @@
 package seedu.address.commons.util;
```
damithc: Use chatGPT to refine the commit message? I spotted some grammar errors.

src/main/java/seedu/address/commons/util/ToStringBuilder.java
```diff
@@ -16,6 +18,7 @@ public class ToStringBuilder {
      * Constructs a {@code ToStringBuilder} whose formatted output will be prefixed with {@code objectName}.
      */
     public ToStringBuilder(String objectName) {
+        requireNonNull(objectName);
         stringBuilder.append(objectName).append(OBJECT_PREFIX);
     }
```
damithc: I don't think we need to check every parameter for null. It need to be used judiciously, to preempt cases where letting a null pass through can cause hard to detect bugs or cause other damage. There is no need to do null checks for cases where a null would cause an obvious and quick failure without affecting system integrity (e.g., without writing incorrect values to the database).
So, we need to think hard if this commit is needed.

src/main/java/seedu/address/model/person/NameContainsKeywordsPredicate.java
```diff
@@ -14,6 +14,9 @@
 public class NameContainsKeywordsPredicate implements Predicate<Person> {
```
damithc: Not related to this PR. Submit it as a separate PR.

damithc (commented):
@Eclipse-Dominator let's include all applicable commits of https://github.com/se-edu/addressbook-level4/pull/973 and not add any other unrelated commits. 
When a PR does a bigger task in multiple smaller steps, we should use a merge commit (not a rebase merge, nor a squash commit) so that both steps and overall outcome are clear -- but we can't do that if there are unrelated commits in the PR.

damithc:
@Eclipse-Dominator how does V2 differ from https://github.com/se-edu/addressbook-level4/pull/973/ ?

Eclipse-Dominator:
> @Eclipse-Dominator how does V2 differ from [se-edu/addressbook-level4#973](https://github.com/se-edu/addressbook-level4/pull/973) ?

It includes some additional standardisation of `equals()` that was not included in ab4#973. The other 2 commit that was left out was from features that was not found in ab3

damithc:
> It includes some additional standardisation of `equals()` that was not included in ab4#973. The other 2 commit that was left out was from features that was not found in ab3

@Eclipse-Dominator I see. Confirmed these additional `equals()` are actually used/needed? We need not add it to very class, as it would add noise without adding value.

Eclipse-Dominator:
> @Eclipse-Dominator I see. Confirmed these additional `equals()` are actually used/needed? We need not add it to very class, as it would add noise without adding value.

There are no additional `equals()` methods added. The first commit just standardizes existing `equal()` implementations. So there is no additional noise added. 

The second commit adds some extra test cases for some of `equals()` method that does not contain a unit test.

docs/tutorials/AddRemark.md
```diff
@@ -118,7 +118,6 @@ public class RemarkCommand extends Command {
```
damithc, line 118: The commit message can be short, as this change doesn't really require a justification. 

```
The style of the #equals(Object) implementation varies
across the code base. Let's standardize it to improve the
consistency of the code.

Co-authored-by: Zhaoqi <e0725180@u.nus.edu>
```

In general, avoid the following when writing commit messages:

* Using `should do X` as if it's a universally accepted rule. Often, it is just our choice, and there could be ways of doing it. Instead, explain it as what the commit does `Let's do X`.
* Reversing the solution and stating it as a problem. e.g., `the code doesn't omit X` (problem) ... `let's omit X` (solution). Often the problem is not an exact opposite of the solution. e.g., Logger is giving a warning when it should not (problem) ... Let's omit X from checkstyle (solution).

damithc (commented):
Just a comment on the commit message. The rest seems good to go.

damithc:

src/main/java/seedu/address/logic/commands/EditCommand.java
```diff
@@ -222,14 +220,12 @@ public boolean equals(Object other) {
...
-            // state check
-            EditPersonDescriptor e = (EditPersonDescriptor) other;
-
-            return getName().equals(e.getName())
-                    && getPhone().equals(e.getPhone())
-                    && getEmail().equals(e.getEmail())
-                    && getAddress().equals(e.getAddress())
-                    && getTags().equals(e.getTags());
+            EditPersonDescriptor otherEditPersonDescriptor = (EditPersonDescriptor) other;
+            return Objects.equals(name, otherEditPersonDescriptor.name)
```
cjyothika, line 224: From my understanding, Objects.equals() considers 2 nulls to be equal. Although, in this case, the name field being null would have been caught already, perhaps it would still be better to use Object.equals()?

cjyothika, line 224: From my understanding, Objects.equals() considers 2 nulls to be equal. Although, in this case, the name field being null would have been caught already, perhaps it would still be better to use Object.equals() or the default equals method of the respective objects?

eg.
`name.equals(otherEditPersonDescriptor.name)`

hansstanley (commented):
Just a suggestion for your consideration (or amusement), perhaps the `#equals` standardization could better follow the DRY principle.

A utility function could be added to `commons.util.AppUtil` (or another shared class):

```java
public class AppUtil {
    ...
    public static <T> boolean isEqual(Class<T> clazz, T target, Object other, BiFunction<T, T, Boolean> resolver) {
        if (target == other) {
            return true;
        }
        if (!clazz.isInstance(other)) { // handles null
            return false;
        }
        return resolver == null || resolver.apply(target, clazz.cast(other));
    }
}
```

Then other classes could use it like so:

```java
public class Phone {
    ...
    @Override
    public boolean equals(Object other) {
        return AppUtil.isEqual(Phone.class, this, other, (p1, p2) -> {
            return p1.value.equals(p2.value);
        });
    }
}
```

Not too sure of any downsides so any criticism is welcome :)

cjyothika (commented):
LGTM. I have one small doubt regarding one of the methods but generally, the changes look good.

damithc:
> Just a suggestion for your consideration (or amusement), perhaps the `#equals` standardization could better follow the DRY principle.
> 
> A utility function could be added to `commons.util.AppUtil` (or another shared class):
> 
> ```java
> public class AppUtil {
>     ...
>     public static <T> boolean isEqual(Class<T> clazz, T target, Object other, BiFunction<T, T, Boolean> resolver) {
>         if (target == other) {
>             return true;
>         }
>         if (!clazz.isInstance(other)) { // handles null
>             return false;
>         }
>         return resolver == null || resolver.apply(target, clazz.cast(other));
>     }
> }
> ```
> 
> Then other classes could use it like so:
> 
> ```java
> public class Phone {
>     ...
>     @Override
>     public boolean equals(Object other) {
>         return AppUtil.isEqual(Phone.class, this, other, (p1, p2) -> {
>             return p1.value.equals(p2.value);
>         });
>     }
> }
> ```
> 
> Not too sure of any downsides so any criticism is welcome :)

Interesting idea @hansstanley I like how it avoids the repetition of the first two checks. One downside I see is that it goes from a very straightforward code to a more complex code. 
Also, the visibly repeating pattern of code might have a learning value in that AB3 developers will remember this implementation as the way to implement the `equals` method (which is not a bad thing).

src/main/java/seedu/address/AppParameters.java
```diff
@@ -50,12 +50,13 @@ public boolean equals(Object other) {
             return true;
         }
 
+        // instanceof handles nulls
```
SPWwj, line 53: This comment might not be necessary.

src/main/java/seedu/address/commons/core/Config.java
```diff
@@ -39,14 +39,15 @@ public boolean equals(Object other) {
         if (other == this) {
             return true;
         }
-        if (!(other instanceof Config)) { //this handles null as well.
+
+        // instanceof handles nulls
```
SPWwj, line 43: Similar to above

src/main/java/seedu/address/logic/commands/EditCommand.java
SPWwj: I don't think this will occur due to requireNonNull(editPersonDescriptor);.

SPWwj (commented):
LGTM, might need to clean up some unnecessary comments.

src/main/java/seedu/address/logic/commands/EditCommand.java
Eclipse-Dominator: it's possible for the local name field to be null. User does not need to edit all the fields of an entry so it's possible to have name being null. In this case `name.equals(...)` will throw nullpointer exception. 

So Objects.equals is used here

src/main/java/seedu/address/AppParameters.java
Eclipse-Dominator: It's to standardize across all equals() so all of the equals() contains this comment. 

`k instanceof clazz` returns false if k is null is not a common way to check for null value of k so the comment is included there.

Eclipse-Dominator:
> 

I think it's quite an interesting solution but I do agree with @damithc that having the current implementation might be better for students viewing the code.

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -69,7 +68,6 @@ public boolean equals(Object other) {
             return false;
         }
 
-        // state check
         PersonCard card = (PersonCard) other;
         return id.getText().equals(card.id.getText())
```
SPWwj: Should we add a comment here to inform the reviewer that the 'Label equals()', the id , behavior differs from our 'equals()' method?

damithc: @Eclipse-Dominator your thoughts?

damithc: > @Eclipse-Dominator your thoughts?

@Eclipse-Dominator reminder ...

Eclipse-Dominator: This method is currently not used in AB3 and does not really fit the idea of `equal`, thus it's better off removed from the source code. 

I have included the commit from AB4 that removes this method in v4

damithc (approved):

PR #198, opened 2023-06-19
Messages: move to seedu.address.logic package

Eclipse-Dominator:
Adapted from: [commit](https://github.com/se-edu/addressbook-level4/pull/974/commits/69a636ff1f7aeef9cc8b458d6b8435b54442d61d)

The Messages class contains common user-facing messages used by commands and parsers in the logic layer.

However, Messages is contained within the seedu.address.commons.core package, which is too general. Given that the Messages class is user-facing, it would only ever be used by the logic and ui layers, and thus would fit better inside the seedu.address.logic package.

Furthermore, in the next few commits we will be adding methods to the Messages class which require access to model classes. This would not be possible if the Messages class is still in the commons layer, which is the lowest layer of our architecture and can't access component.

As such, let's move the Messages class to the seedu.address.logic package.

### DIFF
--- a/src/main/java/seedu/address/logic/Messages.java
+++ b/src/main/java/seedu/address/logic/Messages.java
@@ -1,4 +1,4 @@
-package seedu.address.commons.core;
+package seedu.address.logic;
 
 import seedu.address.model.person.Person;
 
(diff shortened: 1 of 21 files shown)


src/main/java/seedu/address/logic/Messages.java
```diff
@@ -1,4 +1,4 @@
-package seedu.address.commons.core;
```
damithc, line 1:
```
Furthermore, in the next few commits we will be adding methods to the
Messages class which require access to model classes. This would not be
possible if the Messages class is still in the commons layer, which is
the lowest layer of our architecture and can't access component.
```

This above commit message suggests that this commit should be part of the toString PR.
In future, let's keep commits of those AB4 PRs together unless the they don't apply to AB3. There may be good reasons for the commits to be together and in that order, although we might not realize at first.

For now, we have two options:
1. Merge this PR first and the toString PR immediately after.
2. Add this commit to the toString PR and close this one.

damithc (changes requested):

damithc:
Merged with changes to the commit message

PR #199, opened 2023-06-19
Upgrade JavaFX to ver 17

### LINKED ISSUE #108: Use JavaFX 16 for Duke/AB3?

### LINKED ISSUE #193: Window/stage not showing up in KDE

Eclipse-Dominator:
Fixes #193 
Fixes #108 

This is an alternative PR to #196 which explores adding an extra log message to inform the user to ignore the warning message while upgrading to an LTS version.

The application uses JavaFX 11 which is reaching its EOL in September
2023. Thus it is better to upgrade JavaFX to the next LTS version which is version 17.

However, after version 16, JavaFX now throws a warning when JavaFX modules are loaded from the classpath. As there is currently no plan to migrate to using Modules in AB3, let's include an additional log message to inform the users to ignore the warning message.

image
image

### DIFF
--- a/src/main/java/seedu/address/Main.java
+++ b/src/main/java/seedu/address/Main.java
@@ -1,6 +1,9 @@
 package seedu.address;
 
+import java.util.logging.Logger;
+
 import javafx.application.Application;
+import seedu.address.commons.core.LogsCenter;
 
 /**
  * The main entry point to the application.
@@ -19,7 +22,20 @@ import javafx.application.Application;
  * to be the entry point of the application, we avoid this issue.
  */
 public class Main {
+    private static Logger logger = LogsCenter.getLogger(Main.class);
+
     public static void main(String[] args) {
+
+        // As per https://github.com/openjdk/jfx/blob/master/doc-files/release-notes-16.md
+        // JavaFX 16 (or later) runtime logs a warning at startup if JavaFX classes are loaded from
+        // the classpath instead of a module.
+        // Our application does not use Java modules yet. Even if it did, modules are ignored when
+        // packed into a FAT Jar file (as we do), which means this warning will persist even then.
+        // The warning however, can be safely ignored. Thus, the following log informs
+        // the user (if looking at the log output) that the said warning appearing in the log
+        // can be ignored.
+
+        logger.warning("The warning about Unsupported JavaFX configuration below can be ignored.");
         Application.launch(MainApp.class, args);
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ task coverage(type: JacocoReport) {
 
 dependencies {
     String jUnitVersion = '5.4.0'
-    String javaFxVersion = '11.0.2'
+    String javaFxVersion = '17.0.7'
 
     implementation group: 'org.openjfx', name: 'javafx-base', version: javaFxVersion, classifier: 'win'
     implementation group: 'org.openjfx', name: 'javafx-base', version: javaFxVersion, classifier: 'mac'
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -424,7 +424,7 @@
     <module name="MissingJavadocMethodCheck">
       <property name="minLineCount" value="1"/>
       <property name="allowMissingPropertyJavadoc" value="true"/>
-      <property name="ignoreMethodNamesRegex" value="(set.*|get.*)"/>
+      <property name="ignoreMethodNamesRegex" value="(set.*|get.*|main)"/>
     </module>
 
     <!-- Checks that every public class, enumeration and interface has a header comment. -->
--- a/src/main/resources/view/CommandBox.fxml
+++ b/src/main/resources/view/CommandBox.fxml
@@ -3,7 +3,7 @@
 <?import javafx.scene.control.TextField?>
 <?import javafx.scene.layout.StackPane?>
 
-<StackPane styleClass="stack-pane" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1">
+<StackPane styleClass="stack-pane" xmlns="http://javafx.com/javafx/17" xmlns:fx="http://javafx.com/fxml/1">
   <TextField fx:id="commandTextField" onAction="#handleCommandEntered" promptText="Enter command here..."/>
 </StackPane>
 
--- a/src/main/resources/view/HelpWindow.fxml
+++ b/src/main/resources/view/HelpWindow.fxml
@@ -9,7 +9,7 @@
 <?import javafx.scene.layout.HBox?>
 <?import javafx.stage.Stage?>
 
-<fx:root resizable="false" title="Help" type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/11" xmlns:fx="http://javafx.com/fxml/1">
+<fx:root resizable="false" title="Help" type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/17" xmlns:fx="http://javafx.com/fxml/1">
   <icons>
     <Image url="@/images/help_icon.png" />
   </icons>
--- a/src/main/resources/view/MainWindow.fxml
+++ b/src/main/resources/view/MainWindow.fxml
@@ -11,7 +11,7 @@
 <?import javafx.scene.layout.StackPane?>
 <?import javafx.scene.layout.VBox?>
 
-<fx:root type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1"
+<fx:root type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/17" xmlns:fx="http://javafx.com/fxml/1"
          title="Address App" minWidth="450" minHeight="600" onCloseRequest="#handleExit">
   <icons>
     <Image url="@/images/address_book_32.png" />
--- a/src/main/resources/view/PersonListCard.fxml
+++ b/src/main/resources/view/PersonListCard.fxml
@@ -9,7 +9,7 @@
 <?import javafx.scene.layout.Region?>
 <?import javafx.scene.layout.VBox?>
 
-<HBox id="cardPane" fx:id="cardPane" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1">
+<HBox id="cardPane" fx:id="cardPane" xmlns="http://javafx.com/javafx/17" xmlns:fx="http://javafx.com/fxml/1">
   <GridPane HBox.hgrow="ALWAYS">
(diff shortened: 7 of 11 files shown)


damithc:
@Eclipse-Dominator also post a screenshot of how the log output looks.

src/main/java/seedu/address/Main.java
```diff
@@ -19,7 +22,23 @@
  * to be the entry point of the application, we avoid this issue.
  */
 public class Main {
+    private static Logger logger = LogsCenter.getLogger(Main.class);
+
+    /**
+     * Main entry point to the application.
+     */
```
damithc: Seems redundant. Remove?

src/main/java/seedu/address/Main.java
```diff
@@ -19,7 +22,23 @@
...
+        /*
+         * Since JavaFX 16, JavaFX runtime logs a warning if javafx.* modules are loaded from the classpath.
+         * This will result in a warning message whenever the application is launched.
+         * As modules are ignored when packed into a FAT Jar file, this warning will always be shown regardless
+         * of whether the application is using modules or not.
+         *
+         * This warning however, can be safely ignored. Thus, this log is used to inform the user that the warning
+         * can be ignored.
+         * https://github.com/openjdk/jfx/blob/master/doc-files/release-notes-16.md
+         */
```
damithc: Original comment was OK. But rewrote to see if the flow can be refined further.
```suggestion
        /*
         * As per https://github.com/openjdk/jfx/blob/master/doc-files/release-notes-16.md
         * JavaFX 16 (or later) runtime logs a warning at startup if JavaFX classes are loaded from
         * the classpath instead of a module.
         * Our application does not use Java modules yet. Even if it did, modules are ignored when
         * packed into a FAT Jar file (as we do), which means this warning will persist even then.
         * The warning however, can be safely ignored. Thus, the following log informs
         * the user (if looking at the log output) that the said warning appearing in the log 
         * can be ignored.
         */
```

damithc (changes requested):
Just a couple of cosmetic suggestions.

src/main/java/seedu/address/Main.java
Eclipse-Dominator: Removing this seems to cause Checkstyle to throw missing javadoc warning

damithc: I see. But why now? The method was there to begin with, right?

damithc (commented):
Looks good.

Suggested commit message (shorter):
```
Exempt main method from javadoc requirement

Let's update checkstyle.xml to omit the main method from
the javadoc requirement, as the purpose of the main method
is clear.
```

Minor, the checkstyle commit should come before the JavaFX commit?

damithc:

src/main/java/seedu/address/Main.java
```diff
@@ -19,7 +22,20 @@
...
+        /*
+         * As per https://github.com/openjdk/jfx/blob/master/doc-files/release-notes-16.md
+         * JavaFX 16 (or later) runtime logs a warning at startup if JavaFX classes are loaded from
+         * the classpath instead of a module.
+         * Our application does not use Java modules yet. Even if it did, modules are ignored when
+         * packed into a FAT Jar file (as we do), which means this warning will persist even then.
+         * The warning however, can be safely ignored. Thus, the following log informs
+         * the user (if looking at the log output) that the said warning appearing in the log
+         * can be ignored.
+         */
```
damithc: Should this be a `//` comment instead?

chia-yh (commented):
LGTM, however, should the following portion of the `.fxml` files (`HelpWindow.fxml`, etc.) be updated as well?
`xmlns="http://javafx.com/javafx/11"`

dlimyy (commented):
LGTM! Similarly, as mentioned by @chia-yh, I believe that all the `.fxml` should be updated to reflect the latest javafx version.

lywich:
LGTM too! I concur with @chia-yh that it would be prudent to update all the .fxml files.

lywich (commented):
LGTM too! I concur with @chia-yh that it would be prudent to update the `.fxml` files to the latest version.

markusyeo:
LGTM!

In addition to @chia-yh 's catch

I did a quick search of other instances of `http://javafx.com/javafx/` appearing in the fxml files and noticed that there are other namespaces still pointing at `http://javafx.com/javafx/8`.

I think we might need to look at updating all of them to be prudent, or consider removing the version too

jundatan (commented):
LGTM. Other than the updating of  `.fxml` which has already been mentioned by other members.

SPWwj (changes requested):
There's a warning that need to resolve.
image

Additionally, all .fxml files in `resources\view` should be updated to the latest version.

Eclipse-Dominator:
> There's a warning that need to resolve. image
> 
> Additionally, all .fxml files in `resources\view` should be updated to the latest version.

I included a separate info log to tell the users to discard this warning. It seems intellij filters away info logs so I updated the logger to use warning instead and it will appear in intellij as such:

image

chia-yh (commented):
LGTM

SPWwj:
> I included a separate info log to tell the users to discard this warning. It seems intellij filters away info logs so I updated the logger to use warning instead and it will appear in intellij as such:

I'm unsure if we can resolve this warning. This solution look like a last resort.

hansstanley:
Small issue since it's a test file, but for consistency should the `xmlns` and `xmlns:fx` of [`validFileWithFxRoot.fxml`](https://github.com/Eclipse-Dominator/addressbook-level3/blob/master/src/test/resources/view/UiPartTest/validFileWithFxRoot.fxml) be updated too? Currently it does not specify version:

```fxml
<fx:root type="seedu.address.ui.TestFxmlObject" xmlns="http://javafx.com/javafx"
            xmlns:fx="http://javafx.com/fxml">
    <text>Hello World!</text>
</fx:root>
```

Eclipse-Dominator:
> > I included a separate info log to tell the users to discard this warning. It seems intellij filters away info logs so I updated the logger to use warning instead and it will appear in intellij as such:
> 
> I'm unsure if we can resolve this warning. This solution look like a last resort.

The proper solution to resolve the issue is not very clean. It mainly involves migrating to use java modules which incur the following problems:
- Upgrading of `jackson` to a later version (in order to use modules) which leads to changes in behavior of the code when converting `Path` objects to file
- The compiled fat Jar file will ignore the modules so the warning will still show up in the final product even though it might not show up after moving to modules. (currently I have not found a way to prevent this)

Alternatively, we can also manually suppress the warning by changing the JavaFX logger but it might not be ideal since using modules is the supposedly "correct" way of using JavaFX.

damithc:
> Small issue since it's a test file, but for consistency should the `xmlns` and `xmlns:fx` of [`validFileWithFxRoot.fxml`](https://github.com/Eclipse-Dominator/addressbook-level3/blob/master/src/test/resources/view/UiPartTest/validFileWithFxRoot.fxml) be updated too? Currently it does not specify version:
> 
> ```
> <fx:root type="seedu.address.ui.TestFxmlObject" xmlns="http://javafx.com/javafx"
>             xmlns:fx="http://javafx.com/fxml">
>     <text>Hello World!</text>
> </fx:root>
> ```

@Eclipse-Dominator any response to this comment?

SPWwj:
> > > I included a separate info log to tell the users to discard this warning. It seems intellij filters away info logs so I updated the logger to use warning instead and it will appear in intellij as such:

The warnings are issued by the JVM and are emitted before the Java code even starts to run. I think your current solution is effectively handle the situation, as the warning doesn't affect the application.

damithc:
> @Eclipse-Dominator any response to this comment?

@Eclipse-Dominator reminder ...

Eclipse-Dominator:
> > Small issue since it's a test file, but for consistency should the `xmlns` and `xmlns:fx` of [`validFileWithFxRoot.fxml`](https://github.com/Eclipse-Dominator/addressbook-level3/blob/master/src/test/resources/view/UiPartTest/validFileWithFxRoot.fxml) be updated too? Currently it does not specify version:
> > ```
> > <fx:root type="seedu.address.ui.TestFxmlObject" xmlns="http://javafx.com/javafx"
> >             xmlns:fx="http://javafx.com/fxml">
> >     <text>Hello World!</text>
> > </fx:root>
> > ```
> 
> @Eclipse-Dominator any response to this comment?

I have updated the format `xmlns` and `xmlns:fx` to follow the current project versions.

damithc:
> I have updated the format `xmlns` and `xmlns:fx` to follow the current project versions.

Did you post a new version (via CIHR) after this change?

Eclipse-Dominator:
> > I have updated the format `xmlns` and `xmlns:fx` to follow the current project versions.
> 
> Did you post a new version (via CIHR) after this change?

I encountered some issues with CIHR. It appears that
> PR not up to date with base (master) head. Rebase required.

even though the current `upgrade-to-jfx-17` branch is already update to date with master

damithc:
> even though the current `upgrade-to-jfx-17` branch is already update to date with master

@Eclipse-Dominator How did you update it? I suspect CIHR requires the branch to be synced with master via a rebase, not a merge.

damithc:
@Eclipse-Dominator Also, update the master branch of your fork (in case that matters). It seems to be behind the upstream master.

Eclipse-Dominator:
> > even though the current `upgrade-to-jfx-17` branch is already update to date with master
> 
> @Eclipse-Dominator How did you update it? I suspect CIHR requires the branch to be synced with master via a rebase, not a merge.

@damithc It's done via a rebase and I can confirm that the current branch is up to date with upstream's master branch. 
I will try to do some force pushes to see if this issue can be resolved.

image

PR #201, opened 2023-07-27
Update UML diagrams to the new version of PlantUML

### LINKED ISSUE #200: Upgrade UML diagrams to the new version of PlantUML
Seems the new version has some breaking changes, as the MarkBind team found (some info [here](https://github.com/MarkBind/markbind/releases/tag/v5.0.0)) e.g., no need for GraphViz on Windows

Both diagrams and Dev Guide might needs updating.

Eclipse-Dominator:
Fixes #200 

Alternative ver where the newer theme is used instead: #202 

As of PlantUML 1.2023.2, there has been a change in behavior when rendering PlantUML diagrams. This causes some of the PlantUML diagrams to be rendered incorrectly or fail to compile. Some of the diagrams are also being rendered in a different theme due to a change in the default color theme.

Therefore some of the PlantUML code and images need to be updated to ensure functionality as well as consistency. The old color scheme also needs to be applied as some of the manual color changes are based on the older theme.


damithc:
Closing in favor of #202

PR #202, opened 2023-07-27
Upgrade plant uml new theme

### LINKED ISSUE #200: Upgrade UML diagrams to the new version of PlantUML

Eclipse-Dominator:
Fixes #200 

Alternative to #201 where the new default theme is used instead of the old rose theme.

PlantUML(1.2023.2) introduced a change in behavior when rendering
diagrams. This causes some of the diagrams to be rendered incorrectly or
fail to compile. Diagrams are also being rendered in a different theme
due to a change in the default color theme.

Therefore, some of the PlantUML code and images need to be updated to
ensure functionality and consistency.

Let's
 * update PlantUML code to ensure they can be rendered properly
 * update the diagrams to ensure consistency between the PlantUML code
   and the rendered image


damithc:
@Eclipse-Dominator do you see any diagrams that is not optimal under the default theme (compared to the old theme)?

For the following, I'm not sure why the Storage component appears in a different color.
image

But while we are at it, we can actually color each of those two boxes with a very light shade of the main color we use the same components in the architecture diagram.

Eclipse-Dominator:
@damithc I believe there is some rendering logic for Packages that contain classes and Packages that doesn't. The `Model` package above contains a hidden Class but the `Storage` package doesn't. This causes a change in rendering.

> do you see any diagrams that is not optimal under the default theme (compared to the old theme)?

Mainly it would be the UndoRedoState diagrams.
image

The main differences between the 2 themes are:
- Classes are no longer colored
- Borders, arrows and lines are black by default (previously red)
- Change in the color of `note`

damithc:
> The main differences between the 2 themes are:
> 
> * Classes are no longer colored
> * Borders, arrows and lines are black by default (previously red)
> * Change in the color of `note`

Personally, I think the old theme is slightly easier to read. But given that using the defaults is less work, I'm fine to switch to the new theme despite the slight loss of readability.

Eclipse-Dominator:
> Personally, I think the old theme is slightly easier to read. But given that using the defaults is less work, I'm fine to switch to the new theme despite the slight loss of readability.

@damithc It will still be the same amount of work as all the formatting the diagram component to the AB3 style still require the student to import `style.puml`. So there are actually no difference whether we are using the new theme or the old theme

damithc (commented):
Looks good. 
The following part of the commit message can be shortened as it is mostly stating the obvious:
```
Therefore, some of the PlantUML code and images need to be updated to
ensure functionality and consistency.

Let's
 * update PlantUML code to ensure they can be rendered properly
 * update the diagrams to ensure consistency between the PlantUML code
   and the rendered image
```
Something like this?
```
Let's update the PlantUML diagrams accordingly.
```

damithc:

docs/images/ModelClassDiagram.png
dlimyy, line 1: For the new image, it would be good if the association role `filtered` can be shifted to the association pointing from ModelManager to Person(similar to the old image) since it is the ModelManager that contains the filtered list of `Person`.

dlimyy (commented):
Overall, LGTM! Just one minor suggestion

docs/images/ModelClassDiagram.png
Eclipse-Dominator: Added an invisible class to separate the 2 labels.

docs/images/CommitActivityDiagram.png
SPWwj, line 1: The style of this diagram is slightly different. Can we adjust the CSS to make the text and nodes stand out, making them easier to read?

docs/images/LogicClassDiagram.png
SPWwj, line 1: The original diagram had better contrast and was easier to read. Can we adjust the CSS to make it similar to the original diagram?

SPWwj (changes requested):
The PR looks good, as it standardizes the hexadecimal color from 5 to 6 digits. However, due to this change, the style of some diagrams may not look as good as before. It would be better if we could adjust the style to make the diagrams easier to read.

1. I've noticed that previously all labels were bold, but in the new version, they are not. It might be beneficial to restore the bold formatting for better readability.

2. The default color of the lines is not the same as in the previous version, but I think this is acceptable as it doesn't significantly impact the understanding of the diagrams.

docs/images/LogicClassDiagram.png
Eclipse-Dominator: I wasn't able to directly make font thicker but I compensated by bumping the font size slightly to make the font appear clearer.

PR #203, opened 2023-07-28
Add lower level logging statement

### LINKED ISSUE #194: Add lower level logging statements
Suggestion: Add some logging statements lower that INFO level, to illustrate the use of various logging levels.

Eclipse-Dominator:
Fixes #194 

Includes some lower-level logging statements as well as moving some of the existing logging statements to other levels.

### DIFF
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -53,13 +53,12 @@ public class MainApp extends Application {
 
         AppParameters appParameters = AppParameters.parse(getParameters());
         config = initConfig(appParameters.getConfigPath());
+        initLogging(config);
 
         UserPrefsStorage userPrefsStorage = new JsonUserPrefsStorage(config.getUserPrefsFilePath());
         UserPrefs userPrefs = initPrefs(userPrefsStorage);
         AddressBookStorage addressBookStorage = new JsonAddressBookStorage(userPrefs.getAddressBookFilePath());
         storage = new StorageManager(addressBookStorage, userPrefsStorage);
-
-        initLogging(config);
 
         model = initModelManager(storage, userPrefs);
 
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -74,6 +74,8 @@ public class MainApp extends Application {
      * or an empty address book will be used instead if errors occur when reading {@code storage}'s address book.
      */
     private Model initModelManager(Storage storage, ReadOnlyUserPrefs userPrefs) {
+        logger.info("Using data file : " + storage.getAddressBookFilePath());
+
         Optional<ReadOnlyAddressBook> addressBookOptional;
         ReadOnlyAddressBook initialData;
         try {
--- a/src/main/java/seedu/address/logic/parser/AddressBookParser.java
+++ b/src/main/java/seedu/address/logic/parser/AddressBookParser.java
@@ -3,9 +3,11 @@ package seedu.address.logic.parser;
 import static seedu.address.logic.Messages.MESSAGE_INVALID_COMMAND_FORMAT;
 import static seedu.address.logic.Messages.MESSAGE_UNKNOWN_COMMAND;
 
+import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import seedu.address.commons.core.LogsCenter;
 import seedu.address.logic.commands.AddCommand;
 import seedu.address.logic.commands.ClearCommand;
 import seedu.address.logic.commands.Command;
@@ -26,6 +28,7 @@ public class AddressBookParser {
      * Used for initial separation of command word and args.
      */
     private static final Pattern BASIC_COMMAND_FORMAT = Pattern.compile("(?<commandWord>\\S+)(?<arguments>.*)");
+    private static final Logger logger = LogsCenter.getLogger(AddressBookParser.class);
 
     /**
      * Parses user input into command for execution.
@@ -42,6 +45,12 @@ public class AddressBookParser {
 
         final String commandWord = matcher.group("commandWord");
         final String arguments = matcher.group("arguments");
+
+        // Note to developers: Change the log level in config.json to enable lower level (i.e. FINE, FINER and lower)
+        // log messages such as the one below.
+        // Lower level log messages should be used sparingly to minimize noise in the code.
+        logger.fine("Command word: " + commandWord + "; Arguments: " + arguments);
+
         switch (commandWord) {
 
         case AddCommand.COMMAND_WORD:
@@ -69,6 +78,7 @@ public class AddressBookParser {
             return new HelpCommand();
 
         default:
+            logger.finer("ParseException: " + userInput);
             throw new ParseException(MESSAGE_UNKNOWN_COMMAND);
         }
     }


src/main/java/seedu/address/MainApp.java
```diff
@@ -53,14 +53,13 @@ public void init() throws Exception {
```
damithc, line 53: Suggested commit message:
```
In the MainApp class, let's move the call to the initLogging method
closer to the start of the execution start point, to minimize the
possibility of log statements being executed before the desired log
level is set.
```

config/checkstyle/checkstyle.xml
```diff
@@ -425,6 +425,7 @@
       <property name="minLineCount" value="1"/>
       <property name="allowMissingPropertyJavadoc" value="true"/>
       <property name="ignoreMethodNamesRegex" value="(set.*|get.*)"/>
+      <property name="tokens" value="METHOD_DEF, ANNOTATION_FIELD_DEF"/>
```
damithc: Some constructors may not need a javadoc due to being self-explanatory, but this is not always the case, especially when there there are multiple constructors. Also, our coding standard doesn't allow omitting javadocs for constructors.
So, let's not do this.
Besides, it's not related to this PR.

src/main/java/seedu/address/MainApp.java
```diff
@@ -108,11 +110,11 @@ protected Config initConfig(Path configFilePath) {
         configFilePathUsed = Config.DEFAULT_CONFIG_FILE;
 
         if (configFilePath != null) {
-            logger.info("Custom Config file specified " + configFilePath);
+            logger.fine("Custom Config file specified " + configFilePath);
             configFilePathUsed = configFilePath;
         }
 
-        logger.info("Using config file : " + configFilePathUsed);
+        logger.config("Using config file : " + configFilePathUsed);
```
damithc: Let's not use the config level. severe -> warning -> info -> fine -> finer -> finest is much cleaner hierarchy with config jumping in the middle and muddling things. Besides, it seems to be Java specific thing which the other levels has better transferability to other languages.

src/main/java/seedu/address/logic/parser/AddressBookParser.java
```diff
@@ -42,33 +45,45 @@ public Command parseCommand(String userInput) throws ParseException {
 
         final String commandWord = matcher.group("commandWord");
         final String arguments = matcher.group("arguments");
+        logger.fine("Command word: " + commandWord);
+        logger.fine("Arguments: " + arguments);
```
damithc: Combine as one log message.

src/main/java/seedu/address/logic/parser/AddressBookParser.java
```diff
@@ -42,33 +45,45 @@ public Command parseCommand(String userInput) throws ParseException {
...
+
         switch (commandWord) {
 
         case AddCommand.COMMAND_WORD:
+            logger.fine("Add command detected");
```
damithc: Let's remove these. They make this already-long method longer.

src/main/java/seedu/address/logic/parser/AddressBookParser.java
```diff
@@ -42,33 +45,45 @@ public Command parseCommand(String userInput) throws ParseException {
 
         final String commandWord = matcher.group("commandWord");
         final String arguments = matcher.group("arguments");
```
damithc, line 47: Suggested to add something like this:
```java
// Note to developers: Change the log level in .... (config file name) to enable lower level (i.e., FINE and lower)
// log messages such as the one below. 
// Currently, lower level log messages are used sparingly, to minimize noise in the code.
```

src/main/java/seedu/address/logic/parser/AddressBookParser.java
```diff
@@ -42,33 +45,45 @@ public Command parseCommand(String userInput) throws ParseException {
...
         case ExitCommand.COMMAND_WORD:
+            logger.fine("Exit command detected");
             return new ExitCommand();
 
         case HelpCommand.COMMAND_WORD:
+            logger.fine("Help command detected");
             return new HelpCommand();
 
         default:
+            logger.fine("Unknown command detected");
```
damithc:
```suggestion
            logger.finer("This user input caused a ParseException: " + userInput);
```

src/main/java/seedu/address/logic/parser/AddressBookParser.java
```diff
@@ -42,33 +45,45 @@ public Command parseCommand(String userInput) throws ParseException {
...
             throw new ParseException(MESSAGE_UNKNOWN_COMMAND);
         }
     }
```
damithc, line 84: Let's remove all config, fine, and finer log messages in other classes. The two in this class should be enough to illustrate their usage.

src/main/java/seedu/address/MainApp.java
```diff
@@ -74,6 +74,8 @@ public void init() throws Exception {
      * or an empty address book will be used instead if errors occur when reading {@code storage}'s address book.
```
damithc, line 73: Suggested commit message:
```
Add examples of lower-level log messages

Let's add a few examples of lower-level log messages, to illustrate to
developers how they can be used in this codebase.
```

damithc (changes requested):
Some suggestions added. Mainly, to reduce the blast radius of this code update. As it is a low-gain change, let's keep the cost low.

src/main/java/seedu/address/logic/parser/AddressBookParser.java
Eclipse-Dominator: Kept changes for `MainApp` and `AddressBookParser`

src/main/java/seedu/address/MainApp.java
```diff
@@ -142,7 +143,7 @@ protected Config initConfig(Path configFilePath) {
      */
     protected UserPrefs initPrefs(UserPrefsStorage storage) {
         Path prefsFilePath = storage.getUserPrefsFilePath();
-        logger.info("Using preference file : " + prefsFilePath);
+        logger.fine("Using preference file : " + prefsFilePath);
 
         UserPrefs initializedPrefs;
```
damithc: Use fine/finer for info that users shouldn't see but useful for developers. If the log messages in this class are useful for users, use info level- otherwise OK to use the fine level.

src/main/java/seedu/address/MainApp.java
damithc: I missed this earlier. Adjust the commit message subject to match the body.

src/main/java/seedu/address/logic/parser/AddressBookParser.java
```diff
@@ -42,6 +45,12 @@ public Command parseCommand(String userInput) throws ParseException {
 
         final String commandWord = matcher.group("commandWord");
         final String arguments = matcher.group("arguments");
+
+        // Note to developers: Change the log level in .... (config.json) to enable lower level (i.e., FINE and lower)
```
damithc: The dots should be replaced by the filename.

src/main/java/seedu/address/logic/parser/AddressBookParser.java
```diff
@@ -42,6 +45,12 @@ public Command parseCommand(String userInput) throws ParseException {
...
+        // log messages such as the one below.
+        // Currently, lower level log messages are used sparingly, to minimize noise in the code.
+        logger.fine(String.format("Command word: %s ; Arguments: %s", commandWord, arguments));
```
damithc: In some cases `+` (rather than String.format) results in more readable code.

damithc (changes requested):
Looks like my review was stuck in pending state until now.

src/main/java/seedu/address/MainApp.java
Eclipse-Dominator: My opinion on this is that users should only see logs when a new file is created or some error occurred while trying to load the file. When the file is loaded correctly from the path, I feel that the logs here aren't actually useful for the users but more for developers since it's expected to be loaded correctly for the user which was why I opt to use fine for logs that say "Using xxx file: <file path>".

The argument to use info logs might be if the user wants to edit/modify the related files manually so it might be useful to show file path location with info but based on the [ug](https://se-education.org/addressbook-level3/UserGuide.html#editing-the-data-file), so there isn't really a need to display the file locations.

damithc: That is a fair point. But the UG mentions the data file only. 
Also, the log is not for errors/warnings only -- if we wanted that, we should have set the default log level to WARN or ERROR. At INFO level, it should show important activities that are happening.

damithc:

damithc:
@Eclipse-Dominator pls update the PR description to match the current state of the changes done.

src/main/java/seedu/address/MainApp.java
```diff
@@ -74,6 +74,8 @@ public void init() throws Exception {
      * or an empty address book will be used instead if errors occur when reading {@code storage}'s address book.
      */
     private Model initModelManager(Storage storage, ReadOnlyUserPrefs userPrefs) {
+        logger.info("Using data file : " + storage.getAddressBookFilePath());
```
markusyeo, line 76: There looks like there is an extra space here before the colon

Current code:
```logger.info("Using data file : " + storage.getAddressBookFilePath());```

Proposed Change:
```logger.info("Using data file: " + storage.getAddressBookFilePath());```

File: src/main/java/seedu/address/logic/parser/AddressBookParser.java [hunk 14]
```diff
@@ -42,6 +45,12 @@ public Command parseCommand(String userInput) throws ParseException {
 
         final String commandWord = matcher.group("commandWord");
         final String arguments = matcher.group("arguments");
+
+        // Note to developers: Change the log level in config.json to enable lower level (i.e., FINE and lower)
+        // log messages such as the one below.
+        // Currently, lower level log messages are used sparingly, to minimize noise in the code.
+        logger.fine("Command word: " + commandWord + " ; Arguments: " + arguments);
```
hansstanley: Perhaps this comment can be rephrased so it doesn't describe a point in time, as markers like "currently" make code comments harder to maintain.

Instead, it could describe the desired usage, for e.g. "Lower level log messages should be used sparingly to minimise noise in the code."

Also there's a space before the inline semicolon, not sure if that's intended.

```suggestion
        // Lower level log messages should be used sparingly to minimize noise in the code.
        logger.fine("Command word: " + commandWord + "; Arguments: " + arguments);
```

src/main/java/seedu/address/logic/parser/AddressBookParser.java
```diff
@@ -69,6 +78,7 @@ public Command parseCommand(String userInput) throws ParseException {
             return new HelpCommand();
 
         default:
+            logger.finer("This user input caused a ParseException: " + userInput);
```
SPWwj: `This` is redundant, as you have already specified the `userInput`.

```suggestion
            logger.finer("User input caused a ParseException: " + userInput);
```
Or
```suggestion
            logger.finer("ParseException: " + userInput);
```

src/main/java/seedu/address/logic/parser/AddressBookParser.java
damithc: > Instead, it could describe the desired usage, for e.g. "Lower level log messages should be used sparingly to minimise noise in the code."

Good point @hansstanley

src/main/java/seedu/address/logic/parser/AddressBookParser.java
```diff
@@ -42,6 +45,12 @@ public Command parseCommand(String userInput) throws ParseException {
 
         final String commandWord = matcher.group("commandWord");
         final String arguments = matcher.group("arguments");
+
+        // Note to developers: Change the log level in config.json to enable lower level (i.e., FINE and lower)
```
SPWwj:
```suggestion
        // Note to developers: Change the log level in config.json to enable lower level (i.e., FINE, FINER and lower)
```

SPWwj (commented):
LGTM. Just some minor suggestion to make the comments more succinct.

damithc:
@Eclipse-Dominator reminder about this

src/main/java/seedu/address/MainApp.java
Eclipse-Dominator: Hi sorry for the late update, the formatting is like to follow the other similar logging statements found in `MainApp` where a space is included before the colon.

src/main/java/seedu/address/logic/parser/AddressBookParser.java
```diff
@@ -69,6 +78,7 @@ public Command parseCommand(String userInput) throws ParseException {
             return new HelpCommand();
 
         default:
+            logger.finer("ParseException: " + userInput);
```
damithc, line 81: Suppose the user input `blah` caused a ParseException. Compare the following two log messages:

`ParseException: blah`
`This user input cased a ParseException: blah`

I think the latter is more self-explanatory. The former can even mislead the reader to think `blah` _is_ the ParseException. 
What do you think?

damithc (changes requested):

damithc:
@Eclipse-Dominator shall we wrap this up soon?

damithc:
Added these commits to master with minor tweaks.

PR #216, opened 2024-06-14
[#214] Fix misalignment of person names by changing index font

### LINKED ISSUE #214: Vertically misaligned names in Person List
image

baskar-gopinath:
Fixes #214 

**Explanation below image**

image

- **Issue:** Alignment problem with index numbers in PersonListCard due to proportional font.
- **Solution:** Changed the index font to Arial Mono, a monospaced font.
- **Font Load:** Added Arial Mono MT Pro Regular font to the project.
- **CSS Update:** Modified `cell_big_index` style to use Arial Mono.
- **FXML Adjustment:** Updated `HBox` and `Label` elements for consistent alignment.
- **Outcome:** Achieved uniform alignment of index numbers, improving the UI appearance.


damithc:
I can't help wondering if there is an alternative solution that doesn't require using a different font just for the index. 
On a side note, this commit also seems to overwrite the change we did in the previous PR?
Another question is whether this is worth fixing at all. Given that the name alignment will shift again when we move from single digit to double digit anyway.

baskargopinath:
@damithc thats a good point it will become misaligned again when using double digits so i guess it isnt worth fixing as well

also, it didnt overwrite the change, the spacing has to be adjusted again due to the different font

damithc:
Alright. Let's close this PR for now. We can reopen it later if needed. 
Sorry for not raising this point earlier.

PR #219, opened 2024-07-27
[#158] Update Jackson dependencies to the latest version (v2.17.2)

### LINKED ISSUE #158: Jackson Dependency has vulnerable version
The dependencies are shown below:
https://github.com/se-edu/addressbook-level3/blob/928f14660452b529512dfcaa90f459ecdb744dca/build.gradle#L60-L61

The corresponding report regarding the dependencies is given in the below file:
[AB3 Dependency-Check Report.pdf](https://github.com/se-edu/addressbook-level3/files/10830541/AB3.Dependency-Check.Report.pdf)

It is recommended to upgrade to at least 2.9.8. However, an issue regarding [relative path serialization](https://github.com/FasterXML/jackson-databind/issues/1422) is causing some test cases to fail due to the output paths being absolute.

aureliony:
Fixes #158.

To fix the issue with relative path serialization, `ObjectMapper` is modified to use `ToStringSerializer` for the `Path` class, which serializes relative paths correctly.

### DIFF
--- a/src/main/java/seedu/address/commons/util/JsonUtil.java
+++ b/src/main/java/seedu/address/commons/util/JsonUtil.java
@@ -37,7 +37,8 @@ public class JsonUtil {
             .setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY)
             .registerModule(new SimpleModule("SimpleModule")
                     .addSerializer(Level.class, new ToStringSerializer())
-                    .addDeserializer(Level.class, new LevelDeserializer(Level.class)));
+                    .addDeserializer(Level.class, new LevelDeserializer(Level.class))
+                    .addSerializer(Path.class, new ToStringSerializer()));
 
     static <T> void serializeObjectToJsonFile(Path jsonFile, T objectToSerialize) throws IOException {
         FileUtil.writeToFile(jsonFile, toJsonString(objectToSerialize));
--- a/build.gradle
+++ b/build.gradle
@@ -57,8 +57,8 @@ dependencies {
     implementation group: 'org.openjfx', name: 'javafx-graphics', version: javaFxVersion, classifier: 'mac'
     implementation group: 'org.openjfx', name: 'javafx-graphics', version: javaFxVersion, classifier: 'linux'
 
-    implementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.7.0'
-    implementation group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-jsr310', version: '2.7.4'
+    implementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.17.2'
+    implementation group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-jsr310', version: '2.17.2'
 
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: jUnitVersion
 


src/main/java/seedu/address/commons/util/JsonUtil.java
```diff
@@ -37,7 +37,8 @@ public class JsonUtil {
             .setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY)
             .registerModule(new SimpleModule("SimpleModule")
                     .addSerializer(Level.class, new ToStringSerializer())
-                    .addDeserializer(Level.class, new LevelDeserializer(Level.class)));
+                    .addDeserializer(Level.class, new LevelDeserializer(Level.class))
+                    .addSerializer(Path.class, new ToStringSerializer()));
```
itstrueitstrueitsrealitsreal, line 41: Good that this was added, but will there be a need to add a deserialiser for the path.class as well?
```suggestion
                    .addSerializer(Path.class, new ToStringSerializer())
                    .addDeserializer(Path.class, new LevelDeserializer(Path.class)));
```

itstrueitstrueitsrealitsreal (commented):
Other changes LGTM, but not too sure if there is a need to deserialize the newly added `path.class` as well.

src/main/java/seedu/address/commons/util/JsonUtil.java
itstrueitstrueitsrealitsreal, line 41: Would there be a need to deserialise the `Path.class` as well?
```suggestion
                    .addSerializer(Path.class, new ToStringSerializer())
                    .addDeserializer(Path.class, new PathDeserializer(Path.class)));
```

aureliony: By default, Jackson will use `com.fasterxml.jackson.databind.ext.NioPathDeserializer` to deserialize the `Path` class, which we should not override as it is working as expected for both absolute and relative paths.

jyue487:
LGTM. 

However I am not sure whether it is the best practice or not. ObjectMapper maps the file to its absolute path, therefore we have no choice but to teach our objectmapper to serialize the file to their relative paths, which is to add a Path serializer to our objectmapper module. Here comes the question, should we use ToStringSerializer or should we design a PathSerializer? I think both works fine, it's just an issue of good practices.

Otherwise, if the severity issue can be waited, I am not sure whether it is fine to just stick to the old version first.

aureliony:
> LGTM.
> 
> However I am not sure whether it is the best practice or not. ObjectMapper maps the file to its absolute path, therefore we have no choice but to teach our objectmapper to serialize the file to their relative paths, which is to add a Path serializer to our objectmapper module. Here comes the question, should we use ToStringSerializer or should we design a PathSerializer? I think both works fine, it's just an issue of good practices.
> 
> Otherwise, if the severity issue can be waited, I am not sure whether it is fine to just stick to the old version first.

I am in favor of using `ToStringSerializer`. Even if we were to create our own custom serializer, it's gonna use `path.toString()` anyway (unless there's another way to get a string representation of the path?), so it would keep things much simpler. Additionally `ToStringSerializer` is already being used to serialize `java.util.logging.Level`.

hjungwoo01 (commented):
LGTM.

The code correctly adds the Path serializer using ToStringSerializer to handle relative path serialization. The implementation is clean and follows a consistent style. Given the context, using ToStringSerializer for Path seems appropriate. The approach seems correct and should work as intended.

ryanozx:
Thanks for investigating the issue regarding path serialisation—this was a major blocker when upgrading the Jackson dependencies! A minor question: Why are we upgrading specifically to 2.17.2 instead of a slightly older version that does not contain the security vulnerability (which, if I recall correctly, is any version that is at least 2.13)?  Can we guarantee that it is compatible with the existing modules, and are there any tests which can prove that no unintentional regressions have been introduced? Thanks!

baskargopinath:
LGTM, just wondering if we need to test this

Issue #223, opened 2024-10-02
Rebase tutorial-add-remark branch


damithc:
Update add-remark branch by rebasing, to fix issues mentioned in in https://github.com/nus-cs2103-AY2425S1/forum/issues/460
Need to update https://se-education.org/guides/tutorials/ab3AddRemark.html as well

drustanyjt:
Hi prof @damithc I was trying to look into this, but I'm a little confused about the steps to take. For reference here are the [9 commits](https://github.com/se-edu/addressbook-level3/compare/master...tutorial-add-remark) on the `tutorial-add-remark` branch:
![image](https://github.com/user-attachments/assets/895decea-eb48-476d-a868-6a5f4d86ff2f)

Would it be correct to say we want to:
1. rebase `tutorial-add-remark` to be ahead of `master` so that we're using the most updated APIs from the rest of AB3 (in particular the `Person` class utility functions) and have a similar project structure (for the static import statement)
2. this would lead to some commits having outdated API calls. So for each of the 9 commits we manually update any code without losing the main learning point from each commit
3. update URLs in the ab3addRemark guide to reference the new commits and if necessary the code blocks that appear in the guide

damithc:
> Would it be correct to say we want to:
> 
>     1. rebase `tutorial-add-remark` to be ahead of `master` so that we're using the most updated APIs from the rest of AB3 (in particular the `Person` class utility functions) and have a similar project structure (for the static import statement)
> 
>     2. this would lead to some commits having outdated API calls. So for each of the 9 commits we manually update any code without losing the main learning point from each commit
> 
>     3. update URLs in the ab3addRemark guide to reference the new commits and if necessary the code blocks that appear in the guide

Yes. that's correct @drustanyjt

drustanyjt:
Got it, I'll work on a PR for this.

PR #225, opened 2024-10-30
Rebased and updated tutorial add remark

drustanyjt:
PR to address #223.

**TLDR**: The purpose of this PR is to fix the outdated tutorial-add-remark branch which is used in AB3's [Add Remark tutorial](https://se-education.org/guides/tutorials/ab3AddRemark.html). 4 out of 9 of the original commits made use of methods that no longer existed, or had broken import statements due to changing project structures.

### Introduction

This PR rebases commits from the [tutorial-add-remark](https://github.com/se-edu/addressbook-level3/compare/master...tutorial-add-remark) branch, and where necessary updates the code within the individual commit to ensure compatibility with modern AB3 (as of October 2024). By default git's rebase incorporates codes from the base commit (the master branch in this case) that our commits do not modify. So we are left with ensuring that the modified code for the Remark command makes use of files in the base commit properly.

### Changes 

The 9 rebased commit's start from `01190de`, and look something like this:
image
For each of our 9 original commits, it might be easier to refer to this [diff](https://github.com/se-edu/addressbook-level3/compare/master...drustanyjt:AB3-J17:updated-tutorial-add-remark) to understand the code change. Changes made manually (if any) are also briefly listed below per commit.

1. Logic: Implement minimal remark command [original commit](https://github.com/se-edu/addressbook-level3/commit/faa22c38945d1ee722d81aec17b8aa170867a273): nil
3. Logic: Teach the app to accept 'remark' but do nothing [original commit](https://github.com/se-edu/addressbook-level3/commit/35eb7286f18a029d39cb7a29df8f172a001e4fd8): nil
4. Logic: Teach the app to accept 'remark' arguments but still do nothing [original commit](https://github.com/se-edu/addressbook-level3/commit/dc6d5139d08f6403da0ec624ea32bd79a2ae0cbf): fix import statements
5. Model: Add Remark class [original commit](https://github.com/se-edu/addressbook-level3/commit/4516e099699baa9e2d51801bd26f016d812dedcc): nil
6. Ui: Add a placeholder for remark in PersonCard GUI [original commit](https://github.com/se-edu/addressbook-level3/commit/850b78879582f38accb05dd20c245963c65ea599): nil
7. Model: Modify Person to support a Remark field [original commit](https://github.com/se-edu/addressbook-level3/commit/ce998c37e65b92d35c91d28c7822cd139c2c0a5c): fix use of ToStringBuilder and missing test case
8. Storage: Add Remark field to JsonAdaptedPerson class [original commit](https://github.com/se-edu/addressbook-level3/commit/556cbd0e03ff224d7a68afba171ad2eb0ce56bbf): fix json representation of tags
9. Ui: Connect Remark field to PersonCard [original commit](https://github.com/se-edu/addressbook-level3/commit/5b98fee11b6b3f5749b6b943c4f3bd3aa049b692): nil
10. Logic: Implement RemarkCommand execute() logic [original commit](https://github.com/se-edu/addressbook-level3/commit/fac8f3fd855d55831ca0cc73313b5943d49d4d6e): fix import statements

### Verification

To check that the changes work, for each of these commits, I used Gradle's test configuration to see if all test cases passed, and run configuration to see if AB3 could compile and be used, followed by using checkstyle to see if there are any formatting errors. To help review this PR it might be useful to do something similar and check individual commits. For this reason, it would likely not be useful to check the diff presented on GitHub since it's been rebased to be ahead of `master`.

### Follow Up

1. I'm not sure whether this PR should directly override the `tutorial-add-remark` branch. If that's the case, rather than merging based at `tutorial-add-remark` current HEAD, it would probably be better to force `tutorial-add-remark` to be reset to `master`'s HEAD before merging, to avoid any remnants of old code persisting. I think an alternative is to create some other updated branch that branches of `master's` current head.
2. If we override the existing `tutorial-add-remark` branch completely, then this PR should not be merged until the corresponding tutorial has been updated, else the tutorial would link to commits that don't exist. The tutorial links to commits likely also  need not be modified until this PR is finalized, since every rebase would change the commit's hash.

### DIFF
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -26,6 +26,7 @@ import seedu.address.model.person.Email;
 import seedu.address.model.person.Name;
 import seedu.address.model.person.Person;
 import seedu.address.model.person.Phone;
+import seedu.address.model.person.Remark;
 import seedu.address.model.tag.Tag;
 
 /**
@@ -99,9 +100,10 @@ public class EditCommand extends Command {
         Phone updatedPhone = editPersonDescriptor.getPhone().orElse(personToEdit.getPhone());
         Email updatedEmail = editPersonDescriptor.getEmail().orElse(personToEdit.getEmail());
         Address updatedAddress = editPersonDescriptor.getAddress().orElse(personToEdit.getAddress());
+        Remark updatedRemark = personToEdit.getRemark(); // edit command does not allow editing remarks
         Set<Tag> updatedTags = editPersonDescriptor.getTags().orElse(personToEdit.getTags());
 
-        return new Person(updatedName, updatedPhone, updatedEmail, updatedAddress, updatedTags);
+        return new Person(updatedName, updatedPhone, updatedEmail, updatedAddress, updatedRemark, updatedTags);
     }
 
     @Override
--- a/src/main/java/seedu/address/logic/commands/RemarkCommand.java
+++ b/src/main/java/seedu/address/logic/commands/RemarkCommand.java
@@ -1,0 +1,16 @@
+package seedu.address.logic.commands;
+
+import seedu.address.model.Model;
+
+/**
+ * Changes the remark of an existing person in the address book.
+ */
+public class RemarkCommand extends Command {
+
+    public static final String COMMAND_WORD = "remark";
+
+    @Override
+    public CommandResult execute(Model model) {
+        return new CommandResult("Hello from remark");
+    }
+}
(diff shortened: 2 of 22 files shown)


damithc:
Thanks for this PR, @drustanyjt 
We'll try to get this merged before the next semester starts.
BTW, would it be cleaner to PR against the master branch so that only the 9 original commits appear in the PR? That way, we can accept this new branch and delete the old branch.

drustanyjt:
@damithc that makes sense, I've edited the PR to reference master.

damithc:
@drustanyjt Thanks again for this PR 💯 . I've moved your commits to [this branch](https://github.com/se-edu/addressbook-level3/tree/tutorial-add-remark-updated).

PR #239, opened 2025-09-06
build: bump Checkstyle to 10.25.0

arshinsikka:
- Upgraded from Checkstyle 10.2 to 10.25.0
- Maintains compatibility with Java 17
- No code violations introduced by the upgrade
- All existing rules remain compatible

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ repositories {
 }
 
 checkstyle {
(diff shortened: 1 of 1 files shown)


damithc:
@arshinsikka Thanks for this PR.
I realised that there was an earlier PR that upgrades Checkstyle to v11.0.0 and merged that instead.

PR #242, opened 2025-10-25
Use jackson's `TypeReference` to deserialize into generic classes without warnings

### LINKED ISSUE #241: Deserializing into generic classes
With the current implementation, one cannot deserialise into a generic class without suppressing warnings. However, this can be achieved using `TypeReference`, from the `jackson` library (that is already a part of the AB3 dependencies).

For instance, this would not work without warnings or suppressions:
```java
Optional<ArrayList<String>> configOptional = ConfigUtil.readConfig(configFilePathUsed);
```

TheMythologist:
Closes #241

### DIFF
--- a/src/main/java/seedu/address/commons/util/ConfigUtil.java
+++ b/src/main/java/seedu/address/commons/util/ConfigUtil.java
@@ -4,6 +4,8 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Optional;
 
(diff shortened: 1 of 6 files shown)


damithc:
Thanks for this PR, @TheMythologist 
We are likely to look into this after this semester is over. In the meantime, please submit a new version with a detailed commit message -- in AB3, we normally use a detailed commit message with a body (see https://se-education.org/guides/conventions/git.html)

